PR #17, opened 2018-09-26
Only run dropdown collision check if dropdown open

ChrisDill:
- Added static bool dropOpen to track if the dropdown is currently open.
- Fixes bug where dropdown would change when closed if a element that overlaps is pressed.


raysan5:
Thanks for reporting and the PR! I'll take a closer look...

PR #27, opened 2019-02-12
Updated the scroll panel example

Demizdor:
Also fixed a bug with the slider going outside of the scrollbar in `GuiScrollBar()`. Here's a preview of the updated example.
screenshot000


raysan5:
Great addition! Thanks @Demizdor!

PR #28, opened 2019-02-13
Added GuiScrollBar() to GuiListViewEx()

Demizdor:
This is the last widget that has a scrollbar. To be honest i'm not satisfied with how i hacked the scrollbar into this thing, at some point someone (me?!) will have to rewrite it, probably to add `BeginScissorMode()` / `EndScissorMode()` for smooth scrolling. 

For now it works and i'm eager to start work on the text widgets so i'll leave it like that.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -2185,7 +2185,6 @@ RAYGUIDEF bool GuiListViewEx(Rectangle bounds, const char **text, int count, int
 
     float barHeight = bounds.height;
     float minBarHeight = 10;
-    float barPosY = 0;
 
     // Update control
     //--------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


raysan5:
@Demizdor Thank you very much for the effort, I agree this control is quite ugly...

Using scissor could seem an easy solution but just note that it happens after rastering in the graphic pipeline, it means, all vertex sent to the gpu are processed despite not being drawn... also I tried to keep raylib as API agnostic as possible, allowing to use it from other engines just hooking some functions (well, that's the idea...).

Tomorrow I'll merge this PR; today I've been working all day in text drawing redesign to properly integrate rIcons and text alignment support... not finished yet. Just expect a that change by tomorrow.

PR #30, opened 2019-03-05
Proposed implementation for GuiTextBox(), GuiValueBox() and GuiSpinner()

Demizdor:
Requesting code review for `GuiTextBox()`, `GuiValueBox()` and `GuiSpinner()` this is what i have so far.  The controls are usable the only problem i see is with the text view going all over the place after using cut/paste.

Also wrote a example to test these changes, now made complete with lovely icons ;)
screenshot_20190305_172855


raysan5:
Hey @Demizdor! Amazing PR! This is a really big set of changes and I'd need some time to review...

At this moment I shouldn't change `GuiTextBox()`, `GuiValueBox()` and `GuiSpinner()` current signatures, some tools depend on that, they should be implemented in the meantime as `GuiTextBoxEx()`, `GuiValueBoxEx()` and `GuiSpinnerEx()`... Actually, maybe we can just keep that new functionality for `GuiTextBoxEx()`, the other two usually use very short text input...

Also, I see that only one text box will be active at any given time, with it cursor positioning and selectable area, once edit mode is cancelled, no cursor or selectable area will remain... maybe we can use that in our favor and just keep raygui internal global variables: `textCursorPos`, `textSelectStart`, `textSelectEnd`, used across all text-fields, activating them at edit-mode enter, what do you thing?

Demizdor:
Thanks @raysan5 

Yep, i also think it's a good idea to keep `GuiTextBox()`, `GuiValueBox()` and `GuiSpinner()` as they are, well maybe adding COPY/PASTE to them, that will be useful.
Keeping `GuiValueBoxEx()` and `GuiSpinnerEx()` will help the user a lot, i speak from experience. Without the ability to move the cursor modifying a really large number is a pain, you have to erase it then type again and it wastes time, with the ability to move the cursor you can change it quicker. I say this because in the particle editor i'm _slowly_ building i got really annoyed with the spinner boxes and choose to replace them with sliders for this exact reason.

As for making those 3 variables global i like the idea, and it could work but it will mean that everytime we focus on a different textbox we will loose selection and cursor position. And how do we know that we switched textboxes to clear those global variables? 

BTW the textboxes work like that (only one text box active at a time) **only in the example** because i wanted it that way :) please check the code. Normally if you put `N` textboxes they will all be active... it's the same behavior that `GuiTextBox()` always had!

raysan5:
Hey @Demizdor! Did you work further on this? 

Could I integrate those changes and review them or should I better wait for additional PRs?

Demizdor:
Sorry for the late reply @raysan5 i was away most of last week :(. Please don't merge this PR I'll try to push some code this week or the next but it's Easter here so i don't know if I'll be able to but yeah i'm still working on it.

PR #34, opened 2019-06-04
Fixed crash when editing textbox at non-end pos

benbeshara:
I was encountering a crash when editing a textbox by adding more characters than the char[] could handle. While attempting this from the end of the textbox it would simply not enter the characters, adding the new characters from a position before the end of the array causes the error.

This is the fix I've implemented.


raysan5:
@benbeshara Nice! Thanks for the fix! :)

PR #42, opened 2019-07-01
GuiTextBoxGetState() and GuiTextBoxSetState() types fixed

yashrk:
Type of `GuiTextBoxGetState()` and the argument of `GuiTextBoxSetState()` is `GuiTextBoxState`, not `int`.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -397,8 +397,8 @@ RAYGUIDEF int GuiTextBoxGetCursor(void);                                // Get c
 RAYGUIDEF void GuiTextBoxSetSelection(int start, int length);           // Set selection of active textbox
 RAYGUIDEF Vector2 GuiTextBoxGetSelection(void);                         // Get selection of active textbox (x - selection start  y - selection length)
 RAYGUIDEF bool GuiTextBoxIsActive(Rectangle bounds);                    // Returns true if a textbox control with specified `bounds` is the active textbox
-RAYGUIDEF int GuiTextBoxGetState(void);                                 // Get state for the active textbox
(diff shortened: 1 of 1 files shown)


raysan5:
@yashrk What's the difference of using `int` instead of `GuiTextBoxState`? Actually, `GuiTextBoxState` is an `int`... In raylib I always use `int` despite typed enums...

yashrk:
> What's the difference of using int instead of GuiTextBoxState? Actually, GuiTextBoxState is an int... 

Not always:
```
#if defined(RAYGUI_TEXTBOX_EXTENDED)
// Gui text box state data
typedef struct {
    int cursor;      // Cursor position in text
    int start;       // Text start position (from where we begin drawing the text)
    int index;       // Text start index (index inside the text of `start` always in sync)
    int select;      // Marks position of cursor when selection has started
} GuiTextBoxState;
#endif
```

And in the `gui_text_box.c` example we have exactly
```
#define RAYGUI_TEXTBOX_EXTENDED
```

Maybe we should do this?
```
#if defined(RAYGUI_TEXTBOX_EXTENDED)
// Gui text box state data
typedef struct {
    int cursor;      // Cursor position in text
    int start;       // Text start position (from where we begin drawing the text)
    int index;       // Text start index (index inside the text of `start` always in sync)
    int select;      // Marks position of cursor when selection has started
} GuiTextBoxState} ;
#else
typedef int GuiTextBoxState;
#endif
```

raysan5:
@yashrk My bad, I didn't check it properly. Thanks for the fix! :)

PR #44, opened 2019-07-14
Patched raygui to make it both C and C++ compliant

SasLuca:
# Explanation
I encountered a lot of errors and warning when using raygui in my C++ project. 
I modified raygui.h to make it both C and C++ compliant.

The most significant changes that I made are:

1. For curly intialisations I created a macro named `CURLY_INIT` which expands to `StructName { }` in C++ and `(StructName) { }` in C.

2. I had to add a lot of casts to float because in C++ I would get the error "Conversion from 'int' to 'float' requires a narrowing conversion".

I tested by compiling an example in C and one in C++ using the MSVC compiler, but I didn't add my testing setup to the project since I didn't want to mess with it that much.
In my testing setup I also didn't set the warning level manually, I just left it on the default, so there might be more warnings when compiling for C++ on higher warning levels.
I also didn't check for compliance with other compilers, though the changes are made are straightforward enough that I don't think it will be a problem.

# Conclusion
Of course the cost of this change is that from now on when using curly brace initialisation we should use the macro `CURLY_INIT` which is not as nice and considering this is a C project this does add a bit of ugliness/friction to the code for the sake of compatibility. I guess we could give a nicer name to the macro.
Also people would need to be careful since certain implicit conversions are not allowed in C++ and require casts.
It would also mean that the code would need to be tested with both a C and C++ compiler in order to maintain compatibility.
Personally I use this library in a C++ codebase so I am willing to update it to ensure C++ compatibility when people make breaking changes.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -159,6 +159,12 @@
 #define NUM_PROPS_DEFAULT               16      // Number of standard properties
 #define NUM_PROPS_EXTENDED               8      // Number of extended properties
 
+#ifdef __cplusplus
+    #define CURLY_INIT(name) name
+#else
+    #define CURLY_INIT(name) (name)
+#endif
+
 //----------------------------------------------------------------------------------
 // Types and Structures Definition
 // NOTE: Some types are required for RAYGUI_STANDALONE usage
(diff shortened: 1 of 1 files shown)


AregevDev:
Nice work, I also got many warning while compiling raygui. 

Very awesome 👍

raysan5:
Thanks for the improvement! :)

PR #60, opened 2019-11-04
Fixed GuiSpinner defines

Lachee:
GuiSpinner and ValueBox were being disabled when the  RAYGUI_TEXTBOX_EXTENDED was set.
This PR moves GuiSpinner and GuiValueBox outside of the else statement for RAYGUI_TEXTBOX_EXTENDED and moves the GuiTextBox into it.


raysan5:
@Lachee thanks for the review!

PR #64, opened 2020-01-10
:bug: Fixed GuiGrid, GuiTextBoxMult, README (outdated #define for icons)

Yunoinsky:
The guigrid don't deal with the parameter subdivs correctly.

I just found out that the multiline textbox in "raygui.h" not work well because of 'index out of bounds'. The lastLine would always be NULL, so that it would cause access to index -1 of the lastLine .

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,13 @@ Styles can be loaded at runtime using raygui `GuiLoadStyle()` function. Simple a
  
 ## raygui icons: ricons
 
-`raygui` includes a separate module with a set of custom handcrafter icons, ready to be used and integrated in a easy way with `raygui`. This module is called `ricons.h` and can be automatically included just defining `RAYGUI_SUPPORT_RICONS` before including `raygui`. 
+`raygui` includes a separate module with a set of custom handcrafter icons, ready to be used and integrated in a easy way with `raygui`. This module is called `ricons.h` and can be automatically included just defining `RAYGUI_SUPPORT_ICONS` before including `raygui`. 
 
 <img align="right" src="images/raygui_ricons.png">
 
 ```c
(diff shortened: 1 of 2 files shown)


raysan5:
@Yunoinsky Thanks for the review. Excuse me, I can't see the problem with `subdivs`, it worked ok for me... please, could you illustrate the issue with some screenshot?

Yunoinsky:
> @Yunoinsky Thanks for the review. Excuse me, I can't see the problem with `subdivs`, it worked ok for me... please, could you illustrate the issue with some screenshot?

Sorry for that I've not tried this on other platform, but for my mac osx (10.15.2 catalina), the result of 
` GuiGrid((Rectangle){ 80, 80, 300, 300 }, 50.0f, 2)` looks like this:
image

`int linesV = ((int)(bounds.width/spacing) + 1)*subdivs;`  -> when i take width as 600, sapcing as 300 and subdivs as 3, this line will give me 9 vertical lines, but the correct result is 7 (3 thick lines  and 4 thin lines
and
` DrawRectangleRec(RAYGUI_CLITERAL(Rectangle){ bounds.x + spacing*i, bounds.y, 1, bounds.height }, ((i%subdivs) == 0)? Fade(GetColor(GuiGetStyle(DEFAULT, LINE_COLOR)), GRID_COLOR_ALPHA*4) : Fade(GetColor(GuiGetStyle(DEFAULT, LINE_COLOR)), GRID_COLOR_ALPHA));`
will draw the line at `x = spacing * i for each i`, but in fact the spacing should be divided by subdivs
and after this commit : 
image

raysan5:
@Yunoinsky wow! nice catch! thanks for the fix! :)

Issue #61, opened 2020-01-02
Compiler Warnings


smalls12:
Hello,
I've started using raygui and noticed some immediate compile warnings.

In my projects I compile with cmake and set:
`set( CMAKE_CXX_FLAGS " -g -Wall -Wextra -fPIC -std=c++17 " )`

This will produce quite a few warnings which I think can be fixed pretty easily.

Should this be addressed?

Thanks :)

raysan5:
what warning do you get? you can send a PR, feel free! 😄

raysan5:
Just for reference, warning were related to `fread()` not checking for correct data reading. For the moment I'm keeping it this way.

smalls12:
Going to reopen, going to address the other warnings I found.

raysan5:
@smalls12 I'm afraid I'm neither merging this change, sorry.

PR #65, opened 2020-01-19
Fixing -Wmissing-field-initializers Warnings

smalls12:
I only fixed one so far, just to see if I should continue or not.


smalls12:
Contributes to #61

raysan5:
@smalls12 Actually `Font guiFont = { 0 };`  already sets to `0` all structure fields by design.

smalls12:
You are right, the warnings I am seeing are with g++.
gcc seems to handle these warnings differently, as in not presenting a warning because of the valid partial initialization.

PR #70, opened 2020-02-07
Fix parameters and guess at constants

hagna:
So is SLIDER_SIZE supposed to be SLIDER_WIDTH?


raysan5:
@hagna thanks for the review! raygui changed lately and examples were not properly updated...

PR #76, opened 2020-03-10
Added new example for a custom property list control!

Demizdor:
list

Hey @raysan5! Made this thing some time ago and just updated it today to run with current raygui, think it might be useful but i might be wrong so it's ok if you don't merge!

Have a nice day!


raysan5:
Hi @Demizdor! As always, a great addition! Thanks! :)

PR #80, opened 2020-04-16
Fix unknown stack array size

RobLoach:
Compiler expected a static value. Writing 64 directly in the declaration should fix it up.


raysan5:
@RobLoach Thanks! Actually that module was updated recently, it could require some review...

PR #81, opened 2020-04-19
Use raylib Text functions instead of strcmp() and strcpy()

RobLoach:
Use the raylib text functions instead of strcmp and strcpy where applicable...

```
TextIsEqual()
TextCopy()
```

### DIFF
--- a/examples/controls_test_suite/controls_test_suite.c
+++ b/examples/controls_test_suite/controls_test_suite.c
@@ -221,13 +221,13 @@ int main()
                 {
                     // TODO: Validate textInput value and save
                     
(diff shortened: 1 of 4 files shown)


raysan5:
@RobLoach Sorry, it's better to keep using the standard libc functions instead of raylib ones. The reason is because `raygui` can be used as standalone mode, independently of raylib, I tried to minimize raylib dependency of the library and adding more raylib dependency will result in more functions to override for anyone willing to use raygui independently of raylib.

RobLoach:
Sounds good with me!

PR #83, opened 2020-04-24
Update CMakeLists.txt

lokeb:
Remove comma after RAYGUI_HEADERS or make install will not install the headers because RAYGUI_HEADERS is not set. ", " is an unreported syntactic error.


raysan5:
@lokeb thanks for the improvement!

PR #85, opened 2020-04-26
fixed package mode in FindRaylib

lokeb:
Fixed package mode to prevent infinite recursion.


raysan5:
@lokeb Thanks for the review!

PR #88, opened 2020-05-04
Update raygui Rectangle to match raylib Rectangle

RobLoach:
Rectangle in raygui should use a float.


raysan5:
@RobLoach thanks for the review!

Issue #94, opened 2020-06-19
GuiTextBoxMulti can't handle strings that contain no spaces, or that contain space-free sequences wider than bound's width


kamisori:
https://github.com/raysan5/raygui/blob/3e61557f251c5c64ed002081870434e9dbaa7ba7/src/raygui.h#L1769 and following.

If there is no space in `text`, `lastSpacePos` is never set to anything but 0. Because of this index `i` gets reset to 0 and an infinite loop ensues.

raysan5:
@kamisori Can this issue be closed?

PR #95, opened 2020-06-19
Fix for issue #94: infinite loop on space-free sequences in text that…

kamisori:
… are wider than textbox bounds

marked no lastSpacePos as -1.
check for a space-char had to come earlier than check for width overrun.
if no space found in this line but line too wide, wrap sequence and start new line with textWidth of last character.
reset lastSpacePos on every new line.

Also limited currentLine to keyCount to avoid infinite loops alltogether.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -1768,21 +1768,27 @@ bool GuiTextBoxMulti(Rectangle bounds, char *text, int textSize, bool editMode)
 
             // Calculate cursor position considering text
             char oneCharText[2] = { 0 };
-            int lastSpacePos = 0;
+            int lastSpacePos = -1;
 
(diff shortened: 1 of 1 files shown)


kamisori:
There is a bug with the cursor position, I haven't figured out yet why its off by one pixel per line but I'm looking into it

raysan5:
Hi @kamisori! Thank you very much for the improvements! That's great! 😄

kamisori:
It's a nice library, I like to work with it :D 

It seems to be something a little bigger than being off by just a pixel, tweak the style and the cursor is off by more than just a pixel per line. I might be able to fix this but it would probably be best to make a separate pull request for that one.

raysan5:
@kamisori Ok, merging this PR for now.

PR #97, opened 2020-07-03
GuiTextBoxMulti now will also break lines when encountering a newline character

kamisori:


raysan5:
@kamisori Thanks for the review! Merged! :D

PR #114, opened 2020-12-14
Handle CRLF line endings reading fontFileName.

badlydrawnrod:
If you have a style file with CRLF line endings and specify a font file on a system that expects LF line endings, then the font file will fail to load with an error such as...

```
WARNING: FILEIO: [./assets/terminal/Mecha.ttf
] Failed to open file
```

...because it sees the CR as part of the name.

Adding '\r' to the negated scanset for sscanf() fixes the issue.


raysan5:
@badlydrawnrod nice! thanks for the fix!

PR #115, opened 2020-12-26
Fix dll import/export and extern to work like raylib

JeffM2501:
The current raygui code does not correctly define RAYGUIDEF for windows DLLs with correct 'C' externs.
This PR changes the RAYGUIDEF over to work like the one in raylib so it works on windows.


raysan5:
@JeffM2501 Thanks for the review! Actually `raygui` is my next stop this holidays... I'll see if I can release an official version...

PR #116, opened 2020-12-28
Add horizontal scroll

nomi-san:
Just add horizontal scroll (on `Shift`+`Mouse wheel`) for Scroll Panel control 😀


raysan5:
@nomi-san Thanks for the improvement! :)

PR #118, opened 2021-01-10
Fix typecast warnings

JeffM2501:
This PR fixes all the typecast warnings for int to float, float to int, and size_t to int. This should prevent systems that treat these warnings as errors from working.

It was tested by running the controls_test_suite with the changes and everything seemed to look good.


lolleko:
You could also update the examples. Those have a bunch of typecasts missing as well.

raysan5:
I'm still considering some parts of this merge, despite default `Rectangle` provided by raylib consists of 4 `float`, `raygui` could be used with other libraries that just define a `Rectangle` type as 4 `int`.

JeffM2501:
Perhaps add a #def that defaults to float but can be changed for standalone mode builds that defines the Rectangle member type and then internally cast to that?

raysan5:
@JeffM2501 I'm merging it for now, I'll think about possible improvements in a future.

PR #121, opened 2021-02-01
Move include guard endif to end of file

### LINKED ISSUE #120: Include guard issues with C++
```cpp
#if defined(__cplusplus)
extern "C" {            // Prevents name mangling of functions
#endif
```
https://github.com/raysan5/raygui/blob/master/src/raygui.h#L177

is defined inside the "include guard block":

https://github.com/raysan5/raygui/blob/master/src/raygui.h#L130 - https://github.com/raysan5/raygui/blob/master/src/raygui.h#L502
```cpp
#ifndef RAYGUI_H
#define RAYGUI_H
...
#endif
```

The closing brace for `extern C` however is defined outside the guard block
https://github.com/raysan5/raygui/blob/master/src/raygui.h#L3721
```cpp
#if defined(__cplusplus)
}            // Prevents name mangling of functions
#endif
```

=>

```cpp
#ifndef RAYGUI_H
#define RAYGUI_H

#if defined(__cplusplus)
extern "C" {            // Prevents name mangling of functions
#endif

...

#endif // RAYGUI_H

...

#if defined(__cplusplus)
}            // Prevents name mangling of functions
#endif
```
Therefore if you add the raygui header more than once in a c++ project you will get an extraneous brace and will be unable to compile.

Any reason why the include guard block ends on line 502 instead of the end of the file?

lolleko:
Resolves #120

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -499,8 +499,6 @@ RAYGUIDEF void GuiClearIconPixel(int iconId, int x, int y);     // Clear icon pi
 RAYGUIDEF bool GuiCheckIconPixel(int iconId, int x, int y);     // Check icon pixel value
 #endif
 
-#endif // RAYGUI_H
-
 
 /***********************************************************************************
 *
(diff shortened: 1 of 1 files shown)


raysan5:
@lolleko thanks for the review!

PR #122, opened 2021-02-24
Fixed bug with stuck slide bars in scroll_panel example

Wirgiliusz:
Scroll_panel example (https://github.com/raysan5/raygui/tree/master/examples/scroll_panel) uses two slide bars that should allow change of content area dimensions, but they were stuck and either set to the maximum value, or 1 while clicked (see the gif1).

acBwOD9msB

Correcting arguments taken by `GuiSliderBar()` functions that correspond to `panelContentRec` dimensions fixed the bug and now slide bars work as intended (see gif2):.
UCGwmIYDhI

It's kinda weird tho how there was `bool` argument in the `GuiSliderBar()` functions which definition doesn't use any argument of that type - was an older version of this function using `bool` type argument and it was left there causing a bug?

### DIFF
--- a/examples/scroll_panel/gui_scroll_panel.c
+++ b/examples/scroll_panel/gui_scroll_panel.c
@@ -78,8 +78,8 @@ int main()
             
             showContentArea = GuiCheckBox((Rectangle){ 565, 80, 20, 20 }, "SHOW CONTENT AREA", showContentArea);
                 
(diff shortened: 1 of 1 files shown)


raysan5:
@Wirgiliusz Thanks for the fix! Yeah, it was using an old version of the function.

PR #129, opened 2021-04-16
Scroll panel fix

Julianiolo:
### I did 2 things:
1. fixed the bug of the scroll bars disappearing
2. implemented scroll bar side checking for drawing the little rectangle in the corner of the two scrollbars

### About 1.
The problem was that the scroll panel width was set to 0
This happened because the way the width was calculated was as follows (simplified):
```cpp
float panelWidth = ...;
float contentWidth = ...;
int scrollBarWidth = (int)(panelWidth / (int)content.height) * (int)panelWidth;
```
This first calculates the ratio between content and panel and then multiplies that by the panelWidth
Now the problem with that is that the ratio will allways be <0 (because content.height>panelWidth) which means when this ratio is converted to an int it will always be 0 => the slide bar width will be 0

To fix this I simply didnt cast the ratio to an int but the entire result like this:
```cpp
int scrollBarWidth = (int)((panelWidth / (int)content.height) * (int)panelWidth);
```

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -898,7 +898,7 @@ Rectangle GuiScrollPanel(Rectangle bounds, Rectangle content, Vector2 *scroll)
     if (hasHorizontalScrollBar)
     {
         // Change scrollbar slider size to show the diff in size between the content width and the widget width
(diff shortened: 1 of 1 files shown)


raysan5:
@Julianiolo Nice! Thanks for the improvement!

PR #131, opened 2021-04-17
Textbox multi fix

Julianiolo:
### fixed cursor height to be on the line
for some reason the formula for calculating the per line offset was
```cpp
(textSize+textInnerPadding)*linecount
```
no idea why the inner padding was used
### made cursor width be dependent on text size
cursor is now always a rect with the aspect ratio 10:1


raysan5:
@Julianiolo Thanks for the fix, I implemented it manually due to a `GuiTextBoxMulti()` redesign.

PR #134, opened 2021-05-06
fix missing borderwidth

Julianiolo:
I added the dependency on the BORDER_WIDTH style to:
- ColorPicker
- HueSelector
- AlphaSelector
- GuiComboBox
all of them were hardcoded to 1 before


Julianiolo:
Also fixed the text of GuiGroupBox being moved left depending on how high BORDER_WIDTH is

PR #135, opened 2021-05-06
fix cursor

Julianiolo:
This fixes the cursor position beeing too far on the left resulting in it "clipping" into the text.
I also reorganized the coursor.x set code to be in one place.
The last thing I did was make the cursor width be dependent on Text size


raysan5:
This control has been redesigned

Issue #164, opened 2021-11-04
GuiValueBox doesn't do bounds checking on new input


Andidy:
Inside the `if (editMode)` check GuiValueBox either updates the value based on user input, or does bounds checking. It doesn't do both at once, this means that until the GuiValueBox has its editmode disabled, its possible to force values into the box that are out of bounds. See the following code snippet showing that the bounds check is done in the other case of the `if (editMode)` branch:
https://github.com/raysan5/raygui/blob/af51e4ccf7d5099df25ea96f0789c4dce1f0a7d5/src/raygui.h#L2185-L2193

For example if your bounds are 0 and 1, entering 2 will work and the 2 will remain in the GuiValueBox's value variable until the next time GuiValueBox is called.

Andidy:
I fixed this by copying the bounds checking code on lines 2191-2192 and placing it after the `if(valueHasChanged) *value = TextToInteger(textValue);` call on line 2185.

PR #165, opened 2021-11-04
Fixed bounds check in `GuiValueBox()`

Andidy:
Fixed bounds check so that out of bounds values don't persist until the next call to `GuiValueBox()`.
#164 is fixed by this PR.


raysan5:
@Andidy thanks for the review!

Issue #163, opened 2021-11-04
Spinner control doesn't bounds check properly


Andidy:
The spinner control doesn't handle its min and max values properly because the check for if the spinner's value is greater than or less than the bounds happens before the input is received.

Here is the bounds checking, currently on lines 2082-2086:
https://github.com/raysan5/raygui/blob/af51e4ccf7d5099df25ea96f0789c4dce1f0a7d5/src/raygui.h#L2082-L2086
The GuiValueBox also does bounds checking on line 2092:
https://github.com/raysan5/raygui/blob/af51e4ccf7d5099df25ea96f0789c4dce1f0a7d5/src/raygui.h#L2092
But then, after all bounds checking has been done, the increment and decrement buttons are processed on 2101-2107:
https://github.com/raysan5/raygui/blob/af51e4ccf7d5099df25ea96f0789c4dce1f0a7d5/src/raygui.h#L2101-L2107

I'm using the spinner to index into an array, and this causes issues because the indices go past the end of the array.

Andidy:
I solved this locally by moving lines 2101-2107 above line 2082

PR #166, opened 2021-11-04
Fixed Spinner control bounds check error

Andidy:
The bounds check occurred before the `GuiButton`s for the increment and decrement feature of the `Spinner` control which meant using the buttons when the spinner value was at the `minValue` or `maxValue` and incremented or decremented past the min or max would not be resolved until the next time the `Spinner` control was called in the code, likely on the next iteration of a program loop. This caused an array index out of bounds issue in my codebase.

#163 is fixed by this PR.


raysan5:
@Andidy thanks for the review.

PR #169, opened 2021-12-19
Fixed: incompatible type for argument 1 of 'UnloadTexture'

pitpit:
I am encountering following issue when building raygui on Raspbery PI zero (natively or cross-compiling).

image

Here's a fix.

I did not have the chance to test it on another architecture.


raysan5:
@pitpit Thanks for the fix!

PR #171, opened 2021-12-26
Fix typo

aleepo:


raysan5:
@aleepo Good catch! Thanks for the fix! 👍🙂

PR #174, opened 2022-01-19
Fixed implementation signature of GuiScrollPanel

### LINKED ISSUE #173: Addition of text parameter to GuiScrollPanel is not mirrored in the implementation
GuiScrollPanel is now defined as:
https://github.com/raysan5/raygui/blob/042df8f00479a4b64098e3adb3fa121cb0e68843/src/raygui.h#L509
but the header-only implementation has a different declaration:
https://github.com/raysan5/raygui/blob/042df8f00479a4b64098e3adb3fa121cb0e68843/src/raygui.h#L1454

DaJobat:
Added the new text parameter to the signature for the implementation of GuiScrollPanel. 
Fixes #173

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -1451,7 +1451,7 @@ void GuiPanel(Rectangle bounds, const char *text)
 }
 
 // Scroll Panel control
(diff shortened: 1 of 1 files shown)


raysan5:

PR #176, opened 2022-02-01
add text parameter to GuiScrollPanel examples

megagrump:
GuiScrollPanel was changed to take a text parameter, but the examples weren't updated. These two failed to build without the change.


megagrump:
Oops sorry, the second commit got in there by mistake. Should've been another PR.

The rationale for the `guiIconName` change is that when it's only in the implementation part, you can't use those names anywhere but in the file that has the `RAYGUI_IMPLEMENTATION` define. That's fine for single-file projects but not in other cases.

raysan5:
@megagrump Thanks for the review!

PR #177, opened 2022-02-02
Added 'saveFileMode' option to file dialog state

dmbfm:
In the Custom File Dialog example, currently it's hard to use it to create a new file when saving, since if you type a non-existing file in the text area, it will get deleted when you press the "Select" button. 

I have added a `saveFileMode` on the dialog's state so that, when this variable is true, the `fileNameText` won't get cleared for non-existing files. 

Since this is only an example, not sure if this PR is a good fit here.

Thanks!


raysan5:
@dmbfm Thanks for the improvement! It requires some further testing but it looks good.

PR #185, opened 2022-03-15
Fix formatting to help with parsing

IrishBruse:


src/raygui.h
```diff
@@ -257,7 +257,7 @@
         float y;
     } Vector2;
 
-    // Vector3 type                 // -- ConvertHSVtoRGB(), ConvertRGBtoHSV()
```
raysan5, line 260: I would prefer to avoid removing this piece of information, it's important for me to know why this struct is requried

src/raygui.h
```diff
@@ -555,9 +555,8 @@ RAYGUIAPI void GuiClearIconPixel(int iconId, int x, int y);     // Clear icon pi
 RAYGUIAPI bool GuiCheckIconPixel(int iconId, int x, int y);     // Check icon pixel value
 
 #if !defined(RAYGUI_CUSTOM_ICONS)
-//----------------------------------------------------------------------------------
```
raysan5, line 558: I would also like to keep the separators, they are a visual help for me.

PR #186, opened 2022-03-16
Add individual alpha controls

FatherOfEgg:
Made certain functions have an `alpha` parameter. Added macros to go along with it, which uses the `guiAlpha` variable.


raysan5:
@FatherOfEgg I'm afraid I'm not interested in this redesign of the library.

PR #187, opened 2022-03-19
Mac dynlib build instructions

konsumer:
This provides mac build instructions. Tested on OSX 12.3, on M1 hardware, with homebrew 3.4.2


raysan5:
@konsumer Thanks for the improvement.

PR #190, opened 2022-03-28
Add max to GuiProgressBar

boukew99:
Progressbar cannot visualy overflow anymore and returning value has meaning now. I considered adding min, but have not needed it yet.


raysan5:
@boukew99 Thanks! Nice improvement! :D

PR #195, opened 2022-04-06
Store secret text as static variable

ffreling:
When compiling with Zig 0.9.1, the current implementation raises this error:

```
error: expected type '[*c]u8', found '*const [16:0]u8'
note: cast discards const qualifier
```

As it does not allow casting away the const qualifier (rightly so in that case), we need to store the "********" placeholder in a writable variable. Using static storage is merely an optimization to avoid putting a "constant" buffer on the stack.


ffreling:
I wasn't so sure about how to fix this so I tried to have a minimal workaround.

Thanks for merging it.

PR #196, opened 2022-04-09
positioning for guiicons with guiIconScale

CrezyDud:
i used a hardcoded 16, but in other function there was used a hardcoded 8 and stuff,
so this is how it's been done before, should work

this automatically centers the icons when scaled

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -3873,7 +3873,12 @@ static void GuiDrawText(const char *text, Rectangle bounds, int alignment, Color
         if (iconId >= 0)
         {
             // NOTE: We consider icon height, probably different than text size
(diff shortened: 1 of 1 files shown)


raysan5:
@CrezyDud Why are the icons centered when scaled?

CrezyDud:
> @CrezyDud Why are the icons centered when scaled?

so when you make for example a guibutton with text that makes it make an icon
so that the icon is in the middle of the button

raysan5:
@CrezyDud Actually, icons can be used with any kind of text and text could be aligned LEFT, CENTER or RIGHT, icon should be properly aligned depending on text alignment.

CrezyDud:
@raysan5 
did that, vertical alignment is still automatic on scale

raysan5:
I'm reviewing this issue myself.

PR #198, opened 2022-04-29
Bugfix: UB: left shift type int

furudbat:
Hi,
I was Testing my Application with  Address Sanitizer and Undefined Behavior Sanitizer enabled and found this runtime error:

```bash
.../libs/raygui/include/raygui.h:3676:17: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
pc_0x55555983a18e###func_GuiDrawIcon###.../libs/raygui/include/raygui.h###line_3676###obj_(desktop-app+0x42e618e)
pc_0x55555983c983###func_GuiDrawText###.../libs/raygui/include/raygui.h###line_3897###obj_(desktop-app+0x42e8983)
```

```cpp
// Draw selected icon using rectangles pixel-by-pixel
void GuiDrawIcon(int iconId, int posX, int posY, int pixelSize, Color color)
{
    #define BIT_CHECK(a,b) ((a) & (1u<<(b)))

    for (int i = 0, y = 0; i < RAYGUI_ICON_SIZE*RAYGUI_ICON_SIZE/32; i++)
    {
        for (int k = 0; k < 32; k++)
        {
            if (BIT_CHECK(guiIcons[iconId*RAYGUI_ICON_DATA_ELEMENTS + i], k)) ///< **HERE** runtime error
            {
            #if !defined(RAYGUI_STANDALONE)
                DrawRectangle(posX + (k%RAYGUI_ICON_SIZE)*pixelSize, posY + y*pixelSize, pixelSize, pixelSize, color);
            #endif
            }

            if ((k == 15) || (k == 31)) y++;
        }
    }
}
```

Seems like `((a) & (1<<(b)))` causes the problem, found this on [stackoverflow](https://stackoverflow.com/questions/53566029/1-31-cannot-be-represented-by-type-int).

**Note:**
* No Crashes or anything, Icons gets shown as normal.
* OS: Linux
* Compiler: clang
* Tools: clang-tidy
* raygui version: 3.2-dev

ps.
~~The local `#define BIT_CHECK` baffle me a bit :D ... needs to be `#undef`, otherwise this can cause " macro redefined [-Wmacro-redefined]".~~


raysan5:
@furudbat thanks for the fix

PR #202, opened 2022-05-15
gui_file_dialog.h Fix Memory leaks

CrezyDud:
There was a lot of leaking

`SUMMARY: AddressSanitizer: 32360856 byte(s) leaked in 31603 allocation(s).`

(408 bytes of that was X11, so they don't count)

### DIFF
--- a/examples/custom_file_dialog/gui_file_dialog.h
+++ b/examples/custom_file_dialog/gui_file_dialog.h
@@ -1,6 +1,6 @@
 /*******************************************************************************************
 *
-*   FileDialog v1.1 - Modal file dialog to open/save files
+*   FileDialog v1.2 - Modal file dialog to open/save files
 *
 *   MODULE USAGE:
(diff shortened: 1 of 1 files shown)


examples/custom_file_dialog/gui_file_dialog.h
```diff
@@ -47,38 +47,38 @@
...
-    Vector2 size;
+	Vector2 position;
+	Vector2 size;
 
-    bool fileDialogActive;
+	bool fileDialogActive;
 
-    bool dirPathEditMode;
-    char dirPathText[256];
+	bool dirPathEditMode;
```
raysan5, line 55: Please, use 4 spaces instead of TABS

CrezyDud: Done

raysan5:
Reviewed with a simpler solution

CrezyDud:
@raysan5 your solution is 54 changes instead of 27

Issue #203, opened 2022-05-21
GetTextWidth does not make sure a guiFont is set


gulrak:
I came across a SEGV when building an experimental gui with raygui having a CheckBox as my first element.

The implementation of `GetTextWidth` calls
```c++
size = MeasureTextEx(guiFont, text + textIconOffset, (float)GuiGetStyle(DEFAULT, TEXT_SIZE), (float)GuiGetStyle(DEFAULT, TEXT_SPACING));
```
But if neither `GuiGetStyle` or `GuiLoadStyle*` is called before, it will crash as `guiFont` is still zero-initialized. Having an argument that calls `GuiGetStyle` is not helping, as it is not guaranteed that side-effects of a call in a sequence have an impact on elements prior in that sequence, the arguments of `MeasureTextEx` being the sequence.

This lead to a crash when making a gui with `GuiCheckBox()` as the first element, as that widget calls `GetTextWidth`
before any other raygui function. The `controls_test_suite` example has a checkbook as first element, but it calls
`ClearBackground(GetColor(GuiGetStyle(DEFAULT, BACKGROUND_COLOR)));` before, that ensures a `guiFont` is set.

PR #204, opened 2022-05-21
Fix for unsafe call of MeasureTextEx in GetTextWidth #203

gulrak:
Fix for the issue #203, making sure a style is loaded by pulling a `GuiGetStyle` call out of the `MeasureTextEx` argument list.


raysan5:
@gulrak Nice! Thanks for the fix! :D

PR #207, opened 2022-05-28
GuiTextBox: Ensure utf8 dosn't overflow buffer and backspace handles it correctly

gulrak:
Currently when inserting a multibyte character to a `GuiTextBox` that almost reached its `textSize` limit, the end of that buffer could be overwritten, as the code expected every char to be just one byte.

The BACKSPACE handling also would just remove a single byte of a potential multibyte code point leaving invalid utf8.


raysan5:
@gulrak Thanks for the improvement! A really elegant and self-contained code improvement! Great! :D

gulrak:
That was fast! Thanks for the merge! 👍

PR #208, opened 2022-05-28
add distinct icon-only buffer for internal use

### LINKED ISSUE #205: controls_test_suite.c is calling GuiTextInputBox with a wrong signature
The `controls_test_suite.c` is calling `GuiTextInputBox` with a wrong signature, as it has the `textInput` as argument for `textMaxSize`.

The signature is:
`int GuiTextInputBox(Rectangle bounds, const char *title, const char *message, const char *buttons, char *text, int textMaxSize, int *secretViewActive);`
The call is:
`int result = GuiTextInputBox((Rectangle){ GetScreenWidth()/2 - 120, GetScreenHeight()/2 - 60, 240, 140 }, "Save", GuiIconText(ICON_FILE_SAVE, "Save file as..."), "Introduce a save file name", "Ok;Cancel", textInput, NULL);`

The resulting dialog looks strange too:
dialog

### LINKED ISSUE #206: GuiSpinner and GuiTextInputBox can not be combined with GuiIconText
Due to the nature of `GuiIconText`, storing its result in a static buffer, GuiSpinner and GuiTextInputBox can not be used with `GuiIconText` output as `text` argument, as these call `GuiIconText` inside for their icons (`x`,`<`,`>`) before drawing the `text`, resulting in the last drawn icon be redrawn as the text.

This is resulting in the "Save..." popup of the `controls_test_suite.c` showing an irritating `x` as the message.

gulrak:
This change handles `GuiIconText` calls with only an icon on a different buffer to allow the use in `GuiSpinner` and `GuiTextInputBox` to display the icons without overwriting user text given to these widgets. It was the least intrusive way I could think of to fix the issue. It would still be a problem if one uses `GuiSpinner` or `GuiTextInputBox` with only an icon as text, but that would look not great as a user experience anyways I guess.

This change would fix #205 and fix #206 and together with the small change in the `controls_test_suite.c` to match some previous change in raygui, it makes the `controls_test_suite.c` work again.

### DIFF
--- a/examples/controls_test_suite/controls_test_suite.c
+++ b/examples/controls_test_suite/controls_test_suite.c
@@ -219,7 +219,7 @@ int main()
             if (showTextInputBox)
             {
                 DrawRectangle(0, 0, GetScreenWidth(), GetScreenHeight(), Fade(RAYWHITE, 0.8f));
-                int result = GuiTextInputBox((Rectangle){ GetScreenWidth()/2 - 120, GetScreenHeight()/2 - 60, 240, 140 }, "Save", GuiIconText(ICON_FILE_SAVE, "Save file as..."), "Introduce a save file name", "Ok;Cancel", textInput, NULL);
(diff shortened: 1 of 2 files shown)


raysan5:
@gulrak thanks for the review!

PR #210, opened 2022-06-09
GuiTextBox(): allow edits that fit textSize but not visual bounds

gulrak:
This change allows edits in GuiTextBox where the text is still fitting textSize but not the visual bounds. In the previous code this lead to an uneditable text box (example is rGuiStyler with style set to cyber disabling the hex color box).
It works by displaying the text right aligned and character clipped when in edit mode and text is overflowing.

As soon as the edit mode is left, the widget behaves as it did before, without clipping or adapted alignment.


raysan5:
@gulrak Nice! Thanks for the improvement!

PR #211, opened 2022-06-09
GuiTextBoxMulti(): Remove two potential out of bounds crashes

gulrak:
This change will fix `GuiTextBoxMulti` to handle inserting multibyte characters at the edge of textSize and backspace of bad multibyte sequences at the beginning of the buffer without out-of-bounds writes.


raysan5:
@gulrak Thanks! :D

PR #212, opened 2022-06-09
Fix merge bug introduced with #211

gulrak:
PR #211 introduced an error from the PR preparation, this fixes it.


raysan5:
@gulrak Oh! I didn't see that! Thanks for the fix!

PR #213, opened 2022-06-13
Fixed scrollbars not moving when mouse leaves scroll bar bounds

TheTophatDemon:
This pull request fixes a quirk of the scroll bar widget.
Whenever a user is dragging on the scroll bar with the mouse, moving the mouse outside of the "track" will cause the scrolling to stop abruptly. 
The behavior for most GUI applications is to continue the scrolling for as long as the mouse button is held, and it can be disruptive to the user for it not to be this way, since keeping the mouse within the scroll bar bounds can be difficult.

The implemented fix is to keep a global Vector2 "pin" that is set to the position the mouse first clicked on when scrolling starts. Then, a scrollbar can remain focused for as long as the Vector2 is within its range, but releasing the mouse resets the Vector2 to {NAN, NAN} so that it fails all collision checks.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -1160,6 +1160,8 @@ static float guiAlpha = 1.0f;               // Gui element transpacency on drawi
 
 static unsigned int guiIconScale = 1;       // Gui icon default scale (if icons enabled)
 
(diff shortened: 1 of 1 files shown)


raysan5:
I'm afraid I avoid control-specific globals, it's a limitation of imgui. I'm afraid your solution will fail if multiple scrollbars are used in an application.

PR #219, opened 2022-08-05
Add CMake option to build static library

FnControlOption:


raysan5:
@FnControlOption Sorry, `raygui` is a single-file header-only library, no plans to change that.

PR #220, opened 2022-08-05
Add Makefile to build shared library

FnControlOption:


raysan5:
@FnControlOption Thanks for the update but I prefer to keep it simple and avoid the Makefile. Also the provided Makefile is Linux only.

FnControlOption:
Okay. I’m confused though because I wrote and tested this on a Mac.

PR #221, opened 2022-08-18
Added Support for MSVC

be-thomas:
Fixed a few MSVC related issues by adding codes only specific to windows.

Works fine now, tested on Visual Studio 2022
.


raysan5:
@be-thomas I'm afraid there is some problem with this PR, you changed all the coding style. No plans to change it.

PR #231, opened 2022-10-09
Made MessageBox and WindowBox close buttons optional

muit:
Just proposing a small change that made my life a bit easier.
Ideally, this would be done using flags, but that's a much bigger change than just this.

This change also keeps the API intact.

What a message box looks like without the close button:
image

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -502,7 +502,7 @@ RAYGUIAPI void GuiSetStyle(int control, int property, int value);       // Set o
 RAYGUIAPI int GuiGetStyle(int control, int property);                   // Get one style property
 
 // Container/separator controls, useful for controls organization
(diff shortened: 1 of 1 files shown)


raysan5:
@muit Sorry, this change breaks the API and adds an unneeded parameters for most use cases. I'm not merging it.

muit:
I get that. Thats why I suggested flags could be used instead.

I find this particular feature very useful, I did attach an example :) 

So I dont care that much on the method that provides this functionality, rather on the functionality itself.

muit:
How does it beak the API though? No code needs to change, its an optional parameter

raysan5:
@muit I'm afraid optional parameters are not supported in plain C language.

muit:
Ahh! Didnt know that!

Miguel
________________________________

PR #236, opened 2022-10-31
Added Image definition for RAYGUI_STANDALONE

DaJobat:
Added Image struct def, updated Font struct to match raylib.
Image is required for RAYGUI_STANDALONE flag to work Font was out of sync with raylib master


raysan5:

PR #240, opened 2022-11-06
Minor Fixes: Warning and errors when building for strict compilers

JeffM2501:
Make some strings const that are never modified.
Round and don't truncate by casting (comment says it wants to round). 

These fixes help raygui work better when compiled with C++ code.

Errors usually are
```
src/raygui.h: In function ‘void GuiDrawText(const char*, Rectangle, int, Color)’:
src/raygui.h:3940:43: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
 3940 |         const char **lines = GetTextLines(text, &lineCount);
      |                                           ^~~~
      |                                           |
      |                                           const char*
src/raygui.h:3892:27: note:   initializing argument 1 of ‘char** GetTextLines(char*, int*)’
 3892 | char **GetTextLines(char *text, int *count)
      |                     ~~~~~~^~~~
src/raygui.h:3940:42: error: invalid conversion from ‘char**’ to ‘const char**’ [-fpermissive]
 3940 |         const char **lines = GetTextLines(text, &lineCount);
      |                              ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
      |                                          |
      |                                          char**
```

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -3284,8 +3284,8 @@ Vector2 GuiGrid(Rectangle bounds, const char *text, float spacing, int subdivs)
         if (CheckCollisionPointRec(mousePoint, bounds))
         {
             // NOTE: Cell values must be rounded to int
-            currentCell.x = (int)((mousePoint.x - bounds.x)/spacing);
-            currentCell.y = (int)((mousePoint.y - bounds.y)/spacing);
+            currentCell.x = roundf((mousePoint.x - bounds.x)/spacing);
+            currentCell.y = roundf((mousePoint.y - bounds.y)/spacing);
         }
     }
     //--------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Peter0x44:
These warnings also show up in C, but it seems most C++ compilers like to promote them to errors.

src/raygui.h
```diff
@@ -3284,8 +3284,8 @@ Vector2 GuiGrid(Rectangle bounds, const char *text, float spacing, int subdivs)
         if (CheckCollisionPointRec(mousePoint, bounds))
         {
             // NOTE: Cell values must be rounded to int
-            currentCell.x = (int)((mousePoint.x - bounds.x)/spacing);
-            currentCell.y = (int)((mousePoint.y - bounds.y)/spacing);
+            currentCell.x = roundf((mousePoint.x - bounds.x)/spacing);
```
raysan5, line 3287: Afair, those lines have been already changed in the past because of issues when dealing with floats, the final solution was a cast to `int`. I can't remember which projects were affected by this... in any case I'm merging and I'll see if anything brokes in the future.

raysan5:
@JeffM2501 Thanks for the review!

JeffM2501:
> `GuiGrid()` change has been reverted after testing it with `rGuiIcons()`, returned cell was not correct when using `roundf()`.
> 
> Another solution has been provided to avoid related warnings.

It that is what is required for it to work, then the comment is wrong. it does not need to be rounded, it needs to be truncated. `truncf `would do that for you and have the same effect without the double type casting.

PR #241, opened 2022-11-07
Minor : readability improvement in GuiCell

JeffM2501:
Instead of double casting a value to get it to truncate, this PR uses `floorf `to get the smallest whole number for the cell. This improves readability and would work for negative cells if that ever becomes a thing.


raysan5:
@JeffM2501 thanks for the re-review! 👍😄

PR #243, opened 2022-11-08
Fix having an extra bracket

RobLoach:


raysan5:
@RobLoach thanks for the fix!

PR #250, opened 2023-01-08
float-casting fixes in raygui.h in order to support compilation with clang 16

KOLANICH:


raysan5:
@KOLANICH thanks for the review

Issue #255, opened 2023-01-30
Slider-bar-based controls do not respond when mouse out of bounds


raysan5:
Some controls using sliders bars have problems when trying to mouse the slider using mouse; when mouse moves out of slider rectangle bounds, the control stop responding to mouse movement. Affected controls:

 - `GuiScrollPanel()`
 - `GuiSlider()`
 - `GuiSliderBar()`
 - `GuiColorBarAlpha()`
 - `GuiColorBarHue()`

This issue is a bit tricky because it requires some additional state variable to track mouse out-of-control-bounds but still affecting the control state; being a stateless imgui library and trying to minimize function input parameters as much as possible, it's difficult to find a solution to this issue... 

Two possible improvements:

 - When mouse pressed, control bounds get slightly increased to respond despite moving out ouf control bounds.
 - Keep track in two global variables (`lastMousePressedPositionX`, `lastMousePressedPositionY`) where was the las point the mouse was pressed and keep moving the control until X-Y until the mouse is released. Note that this behaviour could interfere with other controls input, but we can maybe assume that risk...

raysan5:
Just tried to implement this feature avoiding additional input parameters and using a single `sliderActive` global variable... and it's not possible. It works but the global variable, as expected, interfers with all the sliders used in the application.

As proposed, and additional control-specific `editMode` input parameter is required to add this functionality. This implies a redesign of the control and a review of all the applications using `raygui` to integrate this functionality; also `rGuiLayout` tool and generated code.

raysan5:
As a temporal solution, an example implementing those custom sliders [has been added ](https://github.com/raysan5/raygui/blob/master/examples/custom_sliders/custom_sliders.c)

PR #256, opened 2023-02-02
Added a new examples : vertical sliders and mouse-owning sliders

odecaux:
These gui elements provide examples on how to extend the library.
They can be dropped as-is into a project, and the "owning" variants of the sliders solve #255 without changing the library API.


raysan5:
@odecaux Thanks for the addition! I will review it more carefully asap!

PR #260, opened 2023-03-06
Have GuiTextBox/GuiTextBoxMulti respect BORDER_WIDTH

shadeops:
This PR allows users to control the BORDER_WIDTH of the text boxes (GuiTextBox and GuiTextBoxMulti) via GuiSetStyle.

Currently, the BORDER_WIDTH of the text boxes is taken into account when in the  'focused' or 'pressed' state. In the 'normal' state this is hard coded to 1. This change replaces the hard coded value and instead does a GuiGetStyle lookup.

By being able to control the TextBox's border we can achieve effects like -
noborder

However currently with the value hard coded to 1 you get this -
border

If this change is counter to Raygui's design or intent please feel free to close.

Thanks for your time.


raysan5:
@shadeops Thanks for the review! :)

PR #263, opened 2023-03-18
Fix Rectangle instanitation with RAYGUI_CLITERAL

RobLoach:
This makes creating a Rectangle use the RAYGUI_CLITERAL to fix some broken Windows builds.


raysan5:
@RobLoach Good catch! Thanks!

PR #269, opened 2023-04-09
missleading doc in gui_curve_edit

smallcluster:
Old comment, _EvalGuiCurve_ doesn't return normalized y anymore.


raysan5:
@smallcluster thanks for the review

PR #270, opened 2023-04-19
Corrected version number and style

hanaxar:
```RAYGUI_VERSION``` was different than the number mentioned in comment section.
Also proposed change applies RayLib versioning style.


raysan5:
Thanks for the review!

PR #271, opened 2023-04-19
Fix error in GuiTabBar

hanaxar:
When ```RAYGUI_NO_ICONS``` defined, getting errors in ```GuiTabBar```. It seems this is copied from ```GuiWindowBox``` and then not edited and forgotten.


raysan5:
@hanaxar Good catch! Thanks for the review!

PR #272, opened 2023-04-19
Avoid compiler warning about signedness

hanaxar:
Proposing to change type of ```textBoxCursorIndex``` from ```unsigned int``` to ```int``` for avoiding bunch of compiler warnings about signedness.
```textBoxCursorIndex``` is used in many arithmethic operations and comparions with int types, the only way to eliminate warnings, is either casting it to int everytime or declaring as int at the beginning. The latter is proposed.


raysan5:
@hanaxar Thanks for the review!

PR #275, opened 2023-04-23
Fix #274 - Adding semicolon

### LINKED ISSUE #274: Missing semi-colon in RAYGUI_DEBUG_TEXT_BOUNDS debug
Thanks for a great set of libraries.   I noticed that when setting RAYGUI_DEBUG_TEXT_BOUNDS then the following is compiled and errors with missing semicolon at the end of the GuiDrawRectangle call

```
#if defined(RAYGUI_DEBUG_TEXT_BOUNDS)
    GuiDrawRectangle(bounds, 0, WHITE, Fade(RED, 0.4f))
#endif
```

andrewharker:
Adding semi colon to fix #274 so RAYGUI_DEBUG_TEXT_BOUNDS can be used.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -4054,7 +4054,7 @@ static void GuiDrawText(const char *text, Rectangle bounds, int alignment, Color
         }
     }
 #if defined(RAYGUI_DEBUG_TEXT_BOUNDS)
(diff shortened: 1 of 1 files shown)


raysan5:
@andrewharker Good catch! Thanks for the fix! 😄

Issue #33, opened 2019-05-29
raygui unusable for C++ projects


sduman:
Hi,

I'm attempting to use raygui with my raylib project, but after generating the project files for VS2019 or Make on OSX, a ton of errors are generated when trying to compile the raygui header.

It looks like there's an issue related to both casting and just the general C vs C++ syntax.

(copied from Visual Studio)
`Error	C2065	'RICONS_SIZE': undeclared identifier	raylib-proj	C:\raylib-proj\include\libs\raygui.h	681	
Error	C2065	'RICONS_SIZE': undeclared identifier	raylib-proj	C:\raylib-proj\include\libs\raygui.h	717	
Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax	raylib-proj	C:\raylib-proj\include\libs\raygui.h	717	
Error	C3861	'DrawIcon': identifier not found	raylib-proj	C:\raylib-proj\include\libs\raygui.h	717	
Error	C2065	'RICONS_SIZE': undeclared identifier	raylib-proj	C:\raylib-proj\include\libs\raygui.h	721	
Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax	raylib-proj	C:\raylib-proj\include\libs\raygui.h	865	
Error	C2065	'RICON_CROSS_SMALL': undeclared identifier	raylib-proj	C:\raylib-proj\include\libs\raygui.h	883	
Error	C2660	'GuiButton': function does not take 1 arguments	raylib-proj	C:\raylib-proj\include\libs\raygui.h	883	
Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax	raylib-proj	C:\raylib-proj\include\libs\raygui.h	909	
Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax	raylib-proj	C:\raylib-proj\include\libs\raygui.h	979	
Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax	raylib-proj	C:\raylib-proj\include\libs\raygui.h	980	
Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax	raylib-proj	C:\raylib-proj\include\libs\raygui.h	1156	
Error	C2397	conversion from 'int' to 'float' requires a narrowing conversion	raylib-proj	C:\raylib-proj\include\libs\raygui.h	1156	
[...]

PR #276, opened 2023-04-29
Fixed C++ compilation in C literal in call to SetShapesTexture

codecat:
This fixes the same problem mentioned in #33, but I guess this particular line was introduced later.

I see there's `RAYGUI_CLITERAL` but also `CLITERAL` from raylib, is there a reason for this distinction?


raysan5:
@codecat thanks for the review!

The distinction us because, despite `raygui` depends a lot on `raylib`, other backends could be  added (`RAYGUI_STANDALONE`) and raylib could be not included...

Issue #278, opened 2023-05-07
Backspace at beginning of non-empty TextBox causes segmentation fault


PROP65:
https://github.com/raysan5/raygui/blob/38a3d100e1dc4ed8762c6b0da85b2638d6ff81de/src/raygui.h#L2230-L2274

KEY_BACKSPACE at the begining of a TextBox causes 'textBoxCursorIndex' on line #2232 to be decremented without checking if it is already 0, causing an invalid address (text + -1) to be passed to 'GetTextWidth' on line #2274.

'GetTextWidth' attempts to read the address causing a segmentation fault on line #3838.

This only happens if the TextBox is not empty.

PR #279, opened 2023-05-07
Fix for #278

PROP65:
Fix for #278


raysan5:
@PROP65 Good catch and thanks for the review!

Issue #277, opened 2023-05-01
Possible Horizontal Scroll bug


BogdanTheGeek:
https://github.com/raysan5/raygui/blob/ab209bc5d18b060f6701e9904c9f3275d8676a81/src/raygui.h#L1661-L1662
The comment specifies Shift + MouseWheel, but the code use `KEY_LEFT_CONTROL` instead of `KEY_LEFT_SHIFT`

raysan5:
@BogdanTheGeek The comment is wrong, feel free to send a PR

raysan5:
@BogdanTheGeek What is the standard for horizontal scrolling in most application? It can be changed to the most common standard.

`raylib` supports horizontal scroll but I tried to minimize `raygui` dependency, also I didn't consider that feature so important, not all users will use a touchpad and it could be done with a keyboard key and mouse that it's quite standard.

Yes, inputs should be configurable but exposing it in a user-friendly way would require a redesign of raygui. No plans for the moment, maybe in a future.

PR #280, opened 2023-05-07
Fix for #277

PROP65:
Fixes incorrect comment from issue #277


raysan5:
@PROP65 thanks for the review

PR #282, opened 2023-05-09
Fix for sliders

hanaxar:
A fix for sliders not dragging outside of bounds.
Added 2 global variables: a state for slider dragging and a rectangle as an identifier for the slider. This way it's possible to fix sliders without breaking changes.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -903,6 +903,10 @@ typedef enum {
     #define RAYGUI_CLITERAL(name) (name)
 #endif
 
+#ifndef RECTANGLE_IS_EQUAL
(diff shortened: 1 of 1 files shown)


raysan5:
@hanaxar Wow! What a solution! Does it really work??? Very impressive! Thanks! I need to test it carefully before merging...

hanaxar:
Yes it seems to work. I tested it with ```scroll_panel.c``` and ```controls_test_suite.c``` examples.

raysan5:
@hanaxar Just tested it and it works perfectly! Thank you very much! This is really a great improvement! 😄

odecaux:
We could probably remove most of examples/custom_sliders and only keep GuiVerticalSlider/Bar/Pro()

raysan5:
@odecaux I prefer to keep them for now, they show another way to implement it not relying on a global variable. It could be useful for some users.

PR #287, opened 2023-05-21
Fixed few stuff about animation curve example

AnstroPleuton:
In C, the struct is not allowed to have default values, as used in examples/animation_curve/gui_curve_edit.h . This PR fixes that and one minor line of code.


raysan5:
@AnstroPleuton Good catch! Thanks for the review! 👍🙂

PR #290, opened 2023-06-03
Update fileName buf size to match buf size ( 64 ) passed to guiTextBox in image_exporter example. 

cabarger:


raysan5:
@cabarger Oh! Good catch! Thanks for the fix! 😄

PR #293, opened 2023-06-19
Added GuiColorPickerHSV() and GuiColorPanelHSV()

Durobot:
This is an alternative implementation of the color picker control.

This implementation receives color value as HSV. It thus does not have to convert the received color value from RGB to HSV and the value picked by the user from HSV to RGB. The conversions are supposed to happen on the user side.
Because of this it avoids hue jittering when the user drags the selector in the color panel.

Here's an example of use and comparison with existing GuiColorPicker(): 
[https://gist.github.com/Durobot/810eacba2669e04c4afd79cb0db8b354](https://gist.github.com/Durobot/810eacba2669e04c4afd79cb0db8b354)

To see the difference, pick a hue between green and blue, or between red and yellow and drag the selector box in the color panel.


raysan5:
@Durobot Thanks for the addition. I try to keep raygui small with limited set of controls, avoiding duplicates/alternative implementations but in this case I'm doing an exception.

PR #296, opened 2023-06-29
Add an API to get the current fade color of the GUI…

anstropleuton:
…(for a specific reason)
The reason being, I want to set the GUI alpha to some value, and draw other things from my code using that value, too, which does not rely on raygui's GUI calls, might be a handy edition.
I believe this actually does get used during development of applications that does not use only raygui, but add their own addition of some codes, too.


raysan5:
@anstropleuton Sorry, I think this function is not required by raygui, user can store alpha on user side if really required.

PR #298, opened 2023-07-09
Fix GuiSlider value pointer

RobLoach:
`GuiSlider()` takes a `float *`, so we must retrieve the address of `value`, rather than just passing it in.

### DIFF
--- a/examples/custom_sliders/custom_sliders.c
+++ b/examples/custom_sliders/custom_sliders.c
@@ -75,7 +75,7 @@ int main()
             // raygui: controls drawing
             //----------------------------------------------------------------------------------
             GuiGroupBox((Rectangle){ 66, 24, 276, 312 }, "STANDARD");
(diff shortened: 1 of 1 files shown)


examples/custom_sliders/custom_sliders.c
```diff
@@ -75,7 +75,7 @@ int main()
             // raygui: controls drawing
             //----------------------------------------------------------------------------------
             GuiGroupBox((Rectangle){ 66, 24, 276, 312 }, "STANDARD");
-            value = GuiSlider((Rectangle){ 96, 48, 216, 16 }, TextFormat("%0.2f", value), NULL, value, 0.0f, 1.0f);
+            value = GuiSlider((Rectangle){ 96, 48, 216, 16 }, TextFormat("%0.2f", value), NULL, &value, 0.0f, 1.0f);
```
raysan5, line 78: Ouch! raylib 4.0 redesign! Actually return value is not `value` anymore, it's intended for state info.

raysan5:
I'm reviewing it...

PR #300, opened 2023-07-11
Make raygui a bit more open

furudbat:
Hi,
I changed the API a bit to use the "private" (static) functions in my other headers.
_(Add `RAYGUIAPI` to the "Module specific Functions")_

## Make Module specific Functions public

I had a problem when using raygui in my custom GUI-Elements Header file, I couldn't use functions like `GuiDrawRectangle` or the (global) variable `guiAlpha`.
_(This only could work when I use raygui with `#define RAYGUI_IMPLEMENTATION`, like in the [custom_sliders.c](https://github.com/raysan5/raygui/blob/master/examples/custom_sliders/custom_sliders.c) example, but my GUI components are split in different header- and cpp-files)_

## add more getters

* add `GuiGetFade`
* add `GuiSliderIsDragging`

## Little Fixes

change `clicked` to `result` in the `RAYGUI_NO_ICONS` case.
```c
#if defined(RAYGUI_NO_ICONS)
    result = GuiButton(closeButtonRec, "x");
#else
    result = GuiButton(closeButtonRec, GuiIconText(ICON_CROSS_SMALL, NULL));
#endif
```

Cast `(void*)` to `(unsigned char*)`, got some pedantic C++ errors.
```c
unsigned char *... = (unsigned char *)RAYGUI_MALLOC(...);
```

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -1497,7 +1497,7 @@ int GuiWindowBox(Rectangle bounds, const char *title)
     GuiSetStyle(BUTTON, BORDER_WIDTH, 1);
     GuiSetStyle(BUTTON, TEXT_ALIGNMENT, TEXT_ALIGN_CENTER);
 #if defined(RAYGUI_NO_ICONS)
-    clicked = GuiButton(closeButtonRec, "x");
+    result = GuiButton(closeButtonRec, "x");
 #else
     result = GuiButton(closeButtonRec, GuiIconText(ICON_CROSS_SMALL, NULL));
 #endif
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -575,6 +575,8 @@ RAYGUIAPI bool GuiIsLocked(void);                               // Check if gui 
 RAYGUIAPI void GuiFade(float alpha);                            // Set gui controls alpha (global state), alpha goes from 0.0f to 1.0f
 RAYGUIAPI void GuiSetState(int state);                          // Set gui state (global state)
 RAYGUIAPI int GuiGetState(void);                                // Get gui state (global state)
+RAYGUIAPI float GuiGetFade(void);                               // Get gui fade (global state)
+RAYGUIAPI bool GuiSliderIsDragging(void);                       // Get gui slider drag (global state)
 
 // Font set/get functions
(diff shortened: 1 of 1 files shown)


raysan5:
@furudbat Sorry, no plans to expose those functions and neither add the proposed ones.

src/raygui.h
```diff
@@ -1491,7 +1497,7 @@ int GuiWindowBox(Rectangle bounds, const char *title)
     GuiSetStyle(BUTTON, BORDER_WIDTH, 1);
     GuiSetStyle(BUTTON, TEXT_ALIGNMENT, TEXT_ALIGN_CENTER);
 #if defined(RAYGUI_NO_ICONS)
-    clicked = GuiButton(closeButtonRec, "x");
+    result = GuiButton(closeButtonRec, "x");
```
raysan5, line 1500: This fix is good, please could you send a PR for this?

src/raygui.h
```diff
@@ -4174,7 +4180,7 @@ static void GuiLoadStyleFromMemory(const unsigned char *fileData, int dataSize)
                 if ((recsDataCompressedSize > 0) && (recsDataCompressedSize != recsDataSize))
                 {
                     // Recs data is compressed, uncompress it
-                    unsigned char *recsDataCompressed = RAYGUI_MALLOC(recsDataCompressedSize);
+                    unsigned char *recsDataCompressed = (unsigned char *)RAYGUI_MALLOC(recsDataCompressedSize);
```
raysan5, line 4183: This fix is good, please could you send a PR for this?

src/raygui.h
```diff
@@ -4214,7 +4220,7 @@ static void GuiLoadStyleFromMemory(const unsigned char *fileData, int dataSize)
                 if ((glyphsDataCompressedSize > 0) && (glyphsDataCompressedSize != glyphsDataSize))
                 {
                     // Glyphs data is compressed, uncompress it
-                    unsigned char *glypsDataCompressed = RAYGUI_MALLOC(glyphsDataCompressedSize);
+                    unsigned char *glypsDataCompressed = (unsigned char *)RAYGUI_MALLOC(glyphsDataCompressedSize);
```
raysan5, line 4223: This fix is good, please could you send a PR for this?

PR #301, opened 2023-07-12
Update raygui.h

manuel5975p:
Cast malloc to unsigned char* for C++


raysan5:
@manuel5975p thanks for the improvement!

PR #302, opened 2023-07-12
Fix window result and pedantic error

furudbat:
change clicked to result in the RAYGUI_NO_ICONS case.

```c
#if defined(RAYGUI_NO_ICONS)
    result = GuiButton(closeButtonRec, "x");
#else
    result = GuiButton(closeButtonRec, GuiIconText(ICON_CROSS_SMALL, NULL));
#endif
```

Cast (void*) to (unsigned char*), and fix pedantic C++ errors.

```c
unsigned char *... = (unsigned char *)RAYGUI_MALLOC(...);
```


raysan5:
@furudbat Thanks for the review!

Issue #303, opened 2023-07-12
Rectangle in ProgressBar shown wrong


furudbat:
I'm updating from 3.6 to 4.x and the progress bar is a bit broken:

**4.x:**
![image](https://github.com/raysan5/raygui/assets/54323028/65392880-dcc7-40f0-8ff2-4918c0a3c2a4)
![image](https://github.com/raysan5/raygui/assets/54323028/da2cafc3-48c2-4e14-a12b-b3340d6336e0)

**3.6:**
![image](https://github.com/raysan5/raygui/assets/54323028/da9b4d90-3766-407a-9e8d-3decc21c61c2)

1. In the older version, the (outer) top-bottom-Rectangle didn't have the same color as the (inner) Rectangle (in blue). 

2. When I change the value to zero, the progress bar left-right-outer-Rectangle breaks:
**controls_test_suite.c**
```c
progressValue = 0;
GuiProgressBar((Rectangle){ 320, 460, 200, 20 }, NULL, TextFormat(" %i%%", (int)(progressValue*100)), &progressValue, 0.0f, 1.0f);
```
![image](https://github.com/raysan5/raygui/assets/54323028/f30a59ab-0618-4b2e-b91f-fe1cd5c02bed)

![image](https://github.com/raysan5/raygui/assets/54323028/d11ea25e-3041-4c64-a106-da1028377fdf)

raysan5:
@furudbat `GuiProgressBar()` has been redesigned for 4.0, to be more visual and different than `GuiSliderBar()`. The drawing issue on 0 value must be addressed...

PR #306, opened 2023-07-19
Fix progressbar border (#303)

furudbat:
image
image
image


raysan5:
@furudbat thanks for the review but I think at 0% is should be just gray, no need to start lighting it if it hasn't actually started moving.

raysan5:
@furudbat Thanks for the review!

Issue #305, opened 2023-07-16
Add Toggle Slider


Arr4y:
As an alternative to checkboxes, there should be support for a toggle slider. It adds more things to play with.

raysan5:
@Arr4y Agree that it could be a nice addition! Here the possible signature:
```c
RAYGUIAPI int GuiToggleSlider(Rectangle bounds, const char *text, bool *active);
```

PR #307, opened 2023-07-19
Add Toggle Slider (#305)

furudbat:
image

### DIFF
--- a/examples/controls_test_suite/controls_test_suite.c
+++ b/examples/controls_test_suite/controls_test_suite.c
@@ -231,8 +231,7 @@ int main()
             GuiListViewEx((Rectangle){ 165, 180, 140, 200 }, listViewExList, 8, &listViewExScrollIndex, &listViewExActive, &listViewExFocus);
 
             //GuiToggle((Rectangle){ 165, 400, 140, 25 }, "#1#ONE", &toggleGroupActive);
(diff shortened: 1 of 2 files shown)


furudbat:
@raysan5 Not sure how to deal with the style for this control.
I'm currently using the `SLIDER` style for the text.

The problem is the `focus` state, it's the same color as the text:
image

raysan5:
@furudbat this is a tricky one, I don't one what's the best design for it, also, it could be a good idea to allow multiple toogle-slide-positions, similar to the `GuiToggleGroup()`

raysan5:
@furudbat I try to minimize the `raygui` API, so a `GuiToggleSlider()` (using the group one) should be enough for most users, no need to add 3 new functions to the API. Users requiring more advance functionality can implement it themselfs.

furudbat:
>  this is a tricky one, I don't one what's the best design for it, also, it could be a good idea to allow multiple toogle-slide-positions, similar to the GuiToggleGroup()

Oh, I think I miss understood you, Users can of cause implement their own `GuiToggleGroup` (see last commit).

For the style, I use the SLIDER properties, for the outer box and for the inner toggle, I used the BUTTON style (cuz it fit the best, IMO).
The text alignment of the inner toggle can be set up with `TOGGLE, TEXT_ALIGNMENT`
_I tried to use the SLIDER properties for the inner toggle and it doesn't look good._

_(Sorry for the mixed up use of the style properties, I was not sure how to add new style properties with all the bells and whistles and default values for a new GUI control)_

raysan5:
@furudbat I'm merging it for further review but not sure if I will keep it.

PR #309, opened 2023-07-24
Added GuiIntBox() and GuiFloatBox() functions

Z0RIK:
GuiIntBox() is basically the same thing as GuiValueBox(), but it works with negative values. Didn't want to delete GuiValueBox() to not cause comaptibility issues with projects that rely upon it. GuiFloatBox() is just a box for float values. It requires TextToFloat() to function properly, i've added implementation for it in RAYGUI_STANDALONE section and created commit to main raylib library (#3195) with same function implementation in rtext.c file.


raysan5:
@Z0RIK I prefer not adding those extra functions to the main library, similar functionality can already be accomplished with provided `GuiValueBox()` and users requiring more specific implementations can implement their own, you can move those functions to a raygui example if you want.

PR #310, opened 2023-07-24
Added example of a calculator app with custom input box

Z0RIK:
Turned my previous pull request into an example, as you suggested.


raysan5:
@Z0RIK Thanks!

PR #311, opened 2023-07-25
Fix ProgressBar borders, when full

furudbat:
This is getting ridiculous, I got off my one pixel (at the right border), and now I hope it's fixed.

**Current**
image

**New:**
image
image


raysan5:
@furudbat Thanks for the review but I'm not merging it because I already fixed it myself some hours ago, I'm doing some other big changes so I will commit everything later on.

PR #314, opened 2023-07-29
Fix C++ build errors

sshsphere:
Add missing `RAYGUI_CLITERAL` and explicit float casts to `GuiDrawRectangle` call


raysan5:
@sshsphere Good catch! Thanks for the fix! 👍🙂

Issue #312, opened 2023-07-26
Custom File dialog directory box


AnkurMal:
Suppose the directory is C:\users\xyz . In the box it shows it correctly. But if I click on a folder inside the directory (lets say it sample), the box shows it as C:\users\xyz/sample instead of  C:\users\xyz\sample. Please look into this.

PR #316, opened 2023-07-29
OS Specific path seperator char

hanaxar:
Addresses #312 and adds OS specific path seperator char.


raysan5:
@hanaxar I got my doubts if this is implementation is the best approach and it will work on all platforms but I'm merging it for now.

Issue #315, opened 2023-07-29
The last letter is not displayed in the GuiMessageBox.


Akimpaus:
The last letter of the message is not displayed in the GuiMessageBox. To make it visible, you have to add a space or some other character at the end of the message. It seems to me that this is not how it was intended...

raysan5:
@Akimpaus Indeed it sounds like an issue, feel free to review it and send a PR! 😄

PR #317, opened 2023-07-29
Solve the issue with last letter of GuiMessageBox

hanaxar:
Solves #315 by adding 2 to ```textWidth```.
All other controls that includes text have 2 pixel extra when measuring text width but somehow it's omitted in ```GuiMessageBox```.


raysan5:
@hanaxar thanks for the review!

PR #320, opened 2023-08-06
minor fix, removed unnecessary float cast for textOffsetY

DaJobat:
textOffsetY is an int, but this addition had a float cast. This was causing issues when using raygui with implementation in zig.


raysan5:

PR #324, opened 2023-08-28
Fix warnings in MSVC

JeffM2501:
A pass to fix warnings in raygui.
One big signed/unsigned thing, and a switch without any cases.

For the unsigned signed issue, I made the global be signed because it's compared signed values.

For the switch, I added in some cases from a comment that just fall through and moved the comment into the case, so at least it'll be easy to add special handling for those cases if/when the time comes, and it gets rid of the warning.


raysan5:
@JeffM2501 thanks for the review!

PR #325, opened 2023-09-01
Added preprocessor macro requirement to readme

CrackedPixel:
This is something that isn't very clear when using the standard header method (the bindings explicitly have it in the example commands)

im not the only one who faced this issue ([example1](https://www.reddit.com/r/raylib/comments/soe1st/how_do_i_use_raygui/?rdt=62800), [example2](https://www.reddit.com/r/raylib/comments/ie5spi/using_raygui/)) and thought something like this may be useful. i wasnt sure where in the readme would be best, so i picked this spot

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Provided set of icons can be reviewed and customized using [rGuiIcons](https://r
 
 ## building
 
-`raygui` is intended to be used as a portable single-file header-only library, to be directly integrated into any C/C++ codebase but some users could require a shared/dynamic version of the library, for example, to create bindings:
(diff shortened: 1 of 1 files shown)


CrackedPixel:
```
*       #define RAYGUI_IMPLEMENTATION
*           Generates the implementation of the library into the included file.
*           If not defined, the library is in header only mode and can be included in other headers
*           or source files without problems. But only ONE file should hold the implementation.
```

i did find this in the .h file, but maybe im not good enough to understand it. i put the macro in my global project settings, not in a specific file

raysan5:
@CrackedPixel I'm afraid that comment could be confusing, the macro is **only required in one compilation unit** but `raygui.h` can be included in multiple files.

CrackedPixel:
> @CrackedPixel I'm afraid that comment could be confusing, the macro is **only required in one compilation unit** but `raygui.h` can be included in multiple files.

That makes sense. would you consider adding some note somewhere in the readme? it would help new users a lot

raysan5:
@CrackedPixel Yes, I can add a note to the readme.

PR #327, opened 2023-09-01
Added optional blinking cursor

CrackedPixel:
This is something that i definitely want in my project, so i took a crack at it (using some of the stuff that was already there and seemingly removed in https://github.com/raysan5/raygui/commit/fdc0ea52497404b6b8661ac641eb0d6b6c96e8c6)

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -562,6 +562,19 @@ typedef enum {
     HUEBAR_SELECTOR_OVERFLOW    // ColorPicker right hue bar selector overflow
 } GuiColorPickerProperty;
 
+// Blink settings
+typedef struct BlinkSettings {
+    bool blinkEnabled;          // true/false if its enabled or not
+    Color blinkColor;           // color of the drawn cursor
+    float width;                // width of the drawn cursor
+    float height;               // height of the drawn cursor
+    int framesShown;            // number of frames cursor will be shown
+    int framesHidden;           // number of frames cursor will be hidden. used to control blink speed
+} BlinkSettings;
+
+static BlinkSettings disabledBlink = { false, { 0x00, 0x00, 0x00, 0xff }, 0.0f, 0.0f, 0, 0 };
+static BlinkSettings defaultBlink = { true, { 0x04, 0x92, 0xc7, 0xff }, 2.0f, 18.0f, 25, 15 };
+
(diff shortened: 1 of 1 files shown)


src/raygui.h
```diff
@@ -562,6 +562,19 @@ typedef enum {
     HUEBAR_SELECTOR_OVERFLOW    // ColorPicker right hue bar selector overflow
 } GuiColorPickerProperty;
 
+// Blink settings
+typedef struct BlinkSettings {
```
CrackedPixel, line 566: per gui element, not global

src/raygui.h
```diff
@@ -2367,6 +2381,13 @@ int GuiTextBox(Rectangle bounds, char *text, int bufferSize, bool editMode)
     // Blink-cursor frame counter
     //if (!autoCursorMode) blinkCursorFrameCounter++;
     //else blinkCursorFrameCounter = 0;
+    if (blinkSettings.blinkEnabled) {
+        if (blinkCursorFrameCounter < (blinkSettings.framesShown + blinkSettings.framesHidden)) {
```
CrackedPixel, line 2385: this will probably cause issues, since multiple elements have their own blinkSettings, but the `blinkCursorFrameCounter` is global

src/raygui.h
```diff
@@ -2480,6 +2501,7 @@ int GuiTextBox(Rectangle bounds, char *text, int bufferSize, bool editMode)
             if (IsKeyPressed(KEY_LEFT) || (IsKeyDown(KEY_LEFT) && (autoCursorCooldownCounter > RAYGUI_TEXTBOX_AUTO_CURSOR_COOLDOWN)))
             {
                 autoCursorDelayCounter++;
+                blinkCursorFrameCounter = 0;
```
CrackedPixel, line 2504: setting it to 0 when left/right is pressed so that as they're moving the cursor it doesnt blink

raysan5:
@CrackedPixel Sorry, no plans to add an additional struct and a function just for a blinking cursor.

Also, considering the common usage for the UI, it could imply not calling it every-frame, only on input-events (I do it with my tools), so blinking cursor won't update as required.

In any case I can consider supporting with current implementation enabling some define flag.

PR #333, opened 2023-09-23
fix: import forthe  example animation_curve

n67094:


raysan5:
@n67094 thanks for the PR! I'm afraid I just reviewed it before seeing it!

PR #336, opened 2023-09-24
Added animation_curve to the examples in CMake

krupitskas:


raysan5:
@krupitskas thanks for the addition

PR #344, opened 2023-10-09
Remove unused params

tobyhinloopen:
My compiler was screaming at me for unused params. I could either disable the warning, or I could remove the unused params. 

Obviously these are breaking changes.


raysan5:
@tobyhinloopen this change is breaking the API and the functions consistency. No plans to change that.

PR #346, opened 2023-10-12
`GuiSliderPro()` returns 1/true if control value has changed

RadsammyT:
This PR proposes a use to the "unused" `int result` seen in `GuiSliderPro()`. `result` will return 1 if the controlled value has been changed, 0 otherwise. 

From what raysan told me in the raylib discord, the results were added in the redesign to return the control's internal states, but they weren't properly designed as those can vary between controls.

I might explore some possible use cases of these return values for other controls later.


raysan5:
@RadsammyT thanks for the addition, I'm still thinking the best way to expose this kind of results, please note that it could be redesigned later on.

Issue #348, opened 2023-11-15
Missing function "DrawLineBezierCubic" on animation_curve example


pulgamecanica:
Hello, I couldn't find anywhere this function:

https://github.com/raysan5/raygui/blob/e765801006820af80933c0ea7bd1ca9834773560/examples/animation_curve/gui_curve_editor.h#L548

I even checked on the raylib headers but it's not there either... can you pls explain me what am I missing? Maybe it comes from somewhere I am not considering.

Also if this is indeed an issue, I would be more than happy to contribute by sharing my solution (I did make it work recreating this missing function)

PR #349, opened 2023-11-15
Implemented new DrawLineBezier Function, (Cuibic) version was deprecated

pulgamecanica:
Hello, while testing your examples module, I realized that you were using the function **`DrawLineBezierCubic`** which was deprecated on raylib5.0

My solution is to implement the **`DrawLineBezier`** function instead.

_Tested with : Linux Fedora 37 x86_64 GNU/Linux_

I hope you find this useful and let me know if there is something else I should consider for this Pull Request.

Solves #348

### DIFF
--- a/examples/animation_curve/gui_curve_editor.h
+++ b/examples/animation_curve/gui_curve_editor.h
@@ -532,20 +532,7 @@ void GuiCurveEditor(GuiCurveEditorState *state, Rectangle bounds)
             {
                 DrawLine(screenPos2.x, screenPos2.y, innerBounds.x+innerBounds.width, screenPos2.y, curveColor);
             }
-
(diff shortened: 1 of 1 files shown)


pulgamecanica:
thanks to Jett (from Discord) for letting me know that the function was indeed deprecated after v5.0

raysan5:
@pulgamecanica Actually the equivalent function is `DrawSplineSegmentBezierCubic()`, `DrawLineBezier()` has no control points.

pulgamecanica:
> @pulgamecanica Actually the equivalent function is `DrawSplineSegmentBezierCubic()`, `DrawLineBezier()` has no control points.

Gotcha, will implement that one then! :)

PR #350, opened 2023-11-16
Implemented new function DrawSplineSegmentBezierCubic to fix example

pulgamecanica:
Hello, while testing your examples, I realized on example: "animation_curve" that you were using the function **`DrawLineBezierCubic`** which was deprecated on raylib5.0

As you explained, the ideal function to use would be **`DrawSplineSegmentBezierCubic`**
I've implemented this function and tested the program to make sure everything is working normally

Tested with : Linux Fedora 37 x86_64 GNU/Linux

I hope you find this useful and let me know if there is something else I should consider for this Pull Request.

Solves #348 

image


raysan5:
@pulgamecanica thanks for the review!

PR #355, opened 2023-12-08
Added GuiValueBoxF for floating point numbers

benjibst:
added a personally needed and requested feature for a value box that works with floating point numbers


benjibst:
tried on linux and windows

PR #356, opened 2023-12-11
Add floating point value box to raygui 

benjibst:


raysan5:
@benjibst this is a breaking change for raygui because it requires a new function from raylib, not merging it for now but I keep the PR open while thinking what should be done with it...

benjibst:
@raysan5 ah ok, got it, maybe one solution could be to just use the c standard library?

raysan5:
> maybe one solution could be to just use the c standard library?

Yeap, probably that's the best solution at this point.

src/raygui.h
```diff
@@ -722,6 +722,7 @@ RAYGUIAPI int GuiComboBox(Rectangle bounds, const char *text, int *active);
 RAYGUIAPI int GuiDropdownBox(Rectangle bounds, const char *text, int *active, bool editMode);          // Dropdown Box control, returns selected item
 RAYGUIAPI int GuiSpinner(Rectangle bounds, const char *text, int *value, int minValue, int maxValue, bool editMode); // Spinner control, returns selected value
 RAYGUIAPI int GuiValueBox(Rectangle bounds, const char *text, int *value, int minValue, int maxValue, bool editMode); // Value Box control, updates input text with numbers
+RAYGUIAPI int GuiValueBoxF(Rectangle bounds,const char* text, char *val_str, float *value, bool editMode); // Floating point Value box control, val_str is buffer where string representation is stored
```
raysan5, line 725: Why is `char *val_str` required? `float *value` should be enough to be equivalent to `GuiValueBox()` function.

benjibst: Hey,
i implemented it this way because when you print out a float without a format specifier you don't know how many digits it will take and when you provide a format specifier for digits after comma, you can never delete a digit because the float is printed with the same number of digits after you delete. For an integer there is only a single way to print it out.

raysan5: I see and it's useful but it would be nice that it worked like `printf()` format parameter so the user could specify something like `03.3` or `2'3` or `4,02`, defining the desired integer-decimal digits, if padded with zeroes and even the format for the decimal separator. Would it be possible to implement that?

Additionally, it should be renamed to `char *valueFormat` and moved after `float *value` parameter.

benjibst: The problem is this: Imagine the user puts in a format specifier %.2f so 2 digits. then, as initial value he gives 100, its printed as 100.00. when you now try to edit the number, say delete a digit, after you delete, the value is parsed(its still 100) and on the next call, its printed out as 100.00 again. so you can never delete a digit

raysan5: But as per my understanding, that's right, depending on the display format edition will require changing more digits from the text but that text is processed after editing

benjibst: ok, i get what you mean, if the text is only parsed after the valuebox loses focus, it should work

benjibst: Hey again, 
after taking a look again i dont think its possible without some sort of text buffer thats persistent between calls. any time the function is called, the float value provided will be converted to text. its impossible to delete the decimal digits because in the next frame they will just be printed out again.

benjibst: > But as per my understanding, that's right, depending on the display format, edition will require changing more digits from the text but that text is processed after editing

the text is always processed while editing, any time you change a digit the text is parsed

raysan5: Then why not use a small static text buffer? `static char text[32]`. It can be only used on editing and avoid processing the float from it until exiting editing mode. Actually, the only limitation of this implementation would be that 2 `valuebox` can not be editable at the same time.

Function can also be called `GuiValueBoxEx()` as an extended value box with float and formatting support (even for integer values).

benjibst: I tried that right now. With a single Valuebox it works fine, but if you have multiple valueboxes the problem is then that every change to the text of one box is overwritten to the other boxes as well, since they have the same text buffer.

raysan5:
@benjibst What is the state of this PR? Is it possible to move `val_str` to a static variable inside the control? It can be updated on editMode and just reseted on exiting editMode. Afaik, in most use cases there won't be two floats being edited at the same time.

benjibst:
@raysan5 last time i tried to implement this with a static text buffer, but the problem is that when you edit the value of one textbox, and then switch to the other textbox, the float value in the text buffer is shared so the value of the second textbox gets overwritten

raysan5:
@benjibst I see the issue. It happens when switching from one text box to another directly without exiting the first one "properly" so it's like two text boxes are active for one frame... mmmh... thinking if it can be addressed in some way...

raysan5:
@benjibst I'm merging this PR and I'll try to find some solution in the meantime... Thanks again for all the work you put on this improvement! :)

benjibst:
ah, do you mean like: if you exit from one valuebox properly, it can parse the string and then clear the buffer? so that the next valuebox knows its allowed to overwrite the static buffer with its own float?

________________________________

raysan5:
@benjibst Yes, that's my idea but I'm not sure if it would work... As per my understanding, at the moment it fails because when you "jump" from one GuiValueBox to another, the state logic for editMode keeps executing without an exit logic... but I could be wrong... it needs to be tested...

PR #358, opened 2023-12-22
round borders and add shadows

felix-lipski:


felix-lipski:
sorry, I selected the wrong target branch

Issue #360, opened 2023-12-30
Change GuiStyleProp propertyValue from int to unsigned int


tyqualters:
VERSION raylib 5.0 / raygui 4.0

raygui.h line 450
```c
typedef struct GuiStyleProp {
    unsigned short controlId;   // Control identifier
    unsigned short propertyId;  // Property identifier
    int propertyValue;          // Property value
} GuiStyleProp;
```

Requesting change `int propertyValue;` to `unsigned int propertyValue;`

```c
typedef struct GuiStyleProp {
    unsigned short controlId;   // Control identifier
    unsigned short propertyId;  // Property identifier
    unsigned int propertyValue;          // Property value
} GuiStyleProp;
```

Reason being that there is a possible narrowing conversion to MAX_INT, and C++ (unlike C) will throw a lot of errors when trying to use any of the default styles (e.g. _/styles/cyber/style_cyber.h_) because of this narrowing conversion. This change fixes that issue and allows C++ to build.

raysan5:
@tyqualters I'm afraid it's not possible to change this type because some properties could have negative values, for example `TEXT_SPACING`.

PR #361, opened 2024-01-02
Fixed narrowing conversion in GuiStyleProp.

Akimpaus:
GUI styles structure uses int, although in most cases it is an unsigned int.
This kind of conversion can cause a warning that stops compilation.
image
This happens when styles are initialized.
image
There are similar conversions in every header style file.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -447,7 +447,7 @@
 typedef struct GuiStyleProp {
     unsigned short controlId;   // Control identifier
     unsigned short propertyId;  // Property identifier
-    int propertyValue;          // Property value
+    unsigned int propertyValue; // Property value
 } GuiStyleProp;
 
 /*
@@ -4844,7 +4844,7 @@ static void GuiDrawText(const char *text, Rectangle textBounds, int alignment, C
             // but we need to draw all of the bad bytes using the '?' symbol moving one byte
             if (codepoint == 0x3f) codepointSize = 1; // TODO: Review not recognized codepoints size
 
-            // Wrap mode text measuring, to validate if 
+            // Wrap mode text measuring, to validate if
             // it can be drawn or a new line is required
             if (wrapMode == TEXT_WRAP_CHAR)
             {


raysan5:
@Akimpaus I'm afraid that change can not be implemented because it will break some properties that require negative values: https://github.com/raysan5/raygui/issues/360

Akimpaus:
@raysan5 But what if, wherever required, explicitly convert to int?

raysan5:
@Akimpaus It would require custom code for some properties at specific moments, I prefer to avoid that. Narrowing conversion errors/warnings should be ignored by compiler in C++. Afaik, that code is valid in C.

Akimpaus:
@raysan5 I mean explicitly convert values to int only in style headers. This would not affect the rest of the code in raygui. Yes, you are right that you can disable these warnings, but, as for me, disabling warnings is not a good practice.

It would be possible to do something like this. All the specified values will be converted from unsigned int to int anyway.
```
static const GuiStyleProp ashesStyleProps[ASHES_STYLE_PROPS_COUNT] = {
    { 0, 0, (int)0xf0f0f0ff },    // DEFAULT_BORDER_COLOR_NORMAL
    { 0, 1, (int)0x868686ff },    // DEFAULT_BASE_COLOR_NORMAL
    { 0, 2, (int)0xe6e6e6ff },    // DEFAULT_TEXT_COLOR_NORMAL
    { 0, 3, (int)0x929999ff },    // DEFAULT_BORDER_COLOR_FOCUSED
    { 0, 4, (int)0xeaeaeaff },    // DEFAULT_BASE_COLOR_FOCUSED
    { 0, 5, (int)0x98a1a8ff },    // DEFAULT_TEXT_COLOR_FOCUSED
    { 0, 6, (int)0x3f3f3fff },    // DEFAULT_BORDER_COLOR_PRESSED
    { 0, 7, (int)0xf6f6f6ff },    // DEFAULT_BASE_COLOR_PRESSED
    { 0, 8, (int)0x414141ff },    // DEFAULT_TEXT_COLOR_PRESSED
    { 0, 9, (int)0x8b8b8bff },    // DEFAULT_BORDER_COLOR_DISABLED
    { 0, 10, (int)0x777777ff },    // DEFAULT_BASE_COLOR_DISABLED
    { 0, 11, (int)0x959595ff },    // DEFAULT_TEXT_COLOR_DISABLED
    { 0, 16, (int)0x00000010 },    // DEFAULT_TEXT_SIZE
    { 0, 18, (int)0x9dadb1ff },    // DEFAULT_LINE_COLOR
    { 0, 19, (int)0x6b6b6bff },    // DEFAULT_BACKGROUND_COLOR
    { 0, 20, (int)0x00000018 },    // DEFAULT_TEXT_LINE_SPACING
};
```

PR #365, opened 2024-01-14
avoid unused parameter wanings

LievenPetersen:
I like to compile my code with `-Werror`, which is impossible due to the unused parameters in some functions.

I see that pr #344 already tried to remove them, which was a bit too extreme.

A better solution, that doesn't change anything, is to write `(void) param;` which avoids the compiler's warning.

This is only about personal preference, without any functional effect really. But it would be really nice to be able to compile with stricter warnings.


raysan5:
@LievenPetersen Personally I prefer to avoid that extra code just to silent some warnings. Also, some of those text variables could be used in the future.

PR #368, opened 2024-01-26
Slider now centers correctly and general UX improvements

jk-:
This addresses two issues I found with the slider.

1. When clicking once in a location, the offset of the slider is not centered.
2. When dragging the slider to the edges, the slider is not compensating for its width.

I also removed an extra branch, did the calculations for the slider value after we calculate everything, and removed the weird (additional) setting of the slider values (:2999-:3000)

Both seen here:
bug

And the fixed version:
fixed

This needs tested with styles that have borders/padding.


raysan5:
@jk- This is really great! It has been on my TODO list for long time! Thank you very much for reviewing it!

PR #374, opened 2024-01-29
gitignore: ignore compiled examples for Linux and OSX

LievenPetersen:
Since executables on Linux (and I believe OSX) don't have a .postfix they were not covered by the other ignores.

Note that I use the * wildcard which does not match arbitrary paths.


raysan5:
@LievenPetersen thanks for the review

PR #375, opened 2024-01-29
fixed and simplified make clean for examples on Linux

LievenPetersen:
The previous command did not work for me (the filters came out empty after grep) and seemed overly complex.


raysan5:
@LievenPetersen thanks for the review!

Issue #369, opened 2024-01-26
`GuiColorBarHue` bug when selecting a color


jk-:
I'm not sure this is the expected behavior? 

The hue bar on the right is moving as you select a color and if you end up on white or black it completely resets

https://github.com/raysan5/raygui/assets/577382/7ed73f84-580f-4eea-9672-aee1919c757e

raysan5:
@jk- I'm afraid this is related to multi-float-operations rounding issues due to the immediate-mode nature of the control. I think https://github.com/raysan5/raygui/pull/370 should minimize this issue...

LievenPetersen:
This happens, because as the RGB Color approaches black or white there is not enough 'color' information to infer a hue. This issue can be avoided by using the HSV equivalent (`GuiColorPickerHSV`), to avoid conversion from RGBA to HSV.

This can be a bit tricky to deal with and basically requires the 'global' color to be stored as HSV and converted to `Color` when needed.

---
### further description of the issue:

The specific sequence of events in `GuiColorPicker` (the Color/RGBA variant) is:
- Color (RGBA) comes in and goes through the `GuiColorPanel` as RGBA
- Then RGB is converted to HSV to go through the `GuiColorBarHue` (this is where the hue value suffers from low resolution)
- Then the hue bar result is converted from HSV to RGB and returned

LievenPetersen:
> @jk- I'm afraid this is related to multi-float-operations rounding issues due to the immediate-mode nature of the control. I think #370 should minimize this issue...

I don't think it will change this. I am not sure that this situation can be improved, because it would require buffering of the hue value, which is goes against the immediate-mode, as you already said.

jk-:
i think docs are fine here. should we update the examples?

LievenPetersen:
> should we update the examples?

I just had a look and it seemed so simple that I just did it :P 
The `controls_test_suite` is the only example that needs updating I think

PR #376, opened 2024-01-29
[Examples] changed GuiColorPicker to HSV variant

LievenPetersen:
As discussed in #369 the color picker was not very stable since the Color variant was used. 

This simple change switches to the HSV variant, which works a lot better.


Callumari08:
Does what it says on the can and bloatless, no reason not to merge!

raysan5:
I prefer to keep the original one.

PR #382, opened 2024-03-13
add floating window example

ssoher:
A reusable floating window function. Window can be moved around through its title bar, minimized with its close button, resized from the icon in the lower right. The window is kept within the application area if the user pushes it outside the window/screen. The contents of the window are clipped with `BeginScissorMode` only if required and also a `GuiScrollPanel` enables user to see the clipped content of the window if current window size does not encapsulate all the content.


raysan5:
@ssoher thanks for the example! I'm merging it and I will review it later more carefully...

ssoher:
Thank you @raysan5, please do review it. I've only recently started using C and I suspect that there are better ways to achieve what I've done in the example. Cheers.

PR #387, opened 2024-03-18
GuiTextBox parameter textSize to bufferSize

SylvainStak:
Not sure if this is called `textSize` in the definition for some specific reason, but in the implementation it's called `bufferSize` (which seems less confusing).


raysan5:
@SylvainStak that's a typo but I prefer to keep name aligned with previous variable, referring to `textSize`.

PR #393, opened 2024-03-30
GuiScrollPanel dynamic mouse scroll speed

SylvainStak:
Test on `scroll_panel.c` example using 100 for bounds width and height.

https://github.com/raysan5/raygui/assets/46388933/7df42e13-e006-4e8e-9cd2-5a12ec985727


raysan5:
@SylvainStak Thanks for the review! Actually, there should be a minimum scroll-bar internal bar size to avoid that almost-invisible bar and speed should be proportional to the content size...

PR #394, opened 2024-04-01
fix compile error in custom input example

emoney17:
and TextToFloat seems to be deprecated so removed it


raysan5:
@emoney17 thanks for the review!

PR #395, opened 2024-04-05
fix typo in raygui.h

Gidan:
Documentation in raygui.h mentions method `GuiLoadStyleDefaulf` which likely should have been `GuiLoadStyleDefault`.


raysan5:
@Gidan Thanks for the review!

PR #396, opened 2024-04-11
Fix division by zero in GuiScrollBar function

Elkantor:
When trying to use the floating window example, I got several times some crashes when expanding the window size.  It's caused in the GuiScrollBar function, when the maxValue is equal to the minValue (I don't know if this case is intended in the first place). Because valueRange variable is maxValue minus minValue, then it is equal to 0, and triggers a crash later, when we use it in division (division by 0).

I'm not sure if it's the best fix, maybe minValue == maxValue is not indentend in the first place and we should fix this problem earlier?

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -5173,7 +5173,9 @@ static int GuiScrollBar(Rectangle bounds, int value, int minValue, int maxValue)
     if (value > maxValue) value = maxValue;
     if (value < minValue) value = minValue;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Elkantor thanks for the review.

 > I'm not sure if it's the best fix, maybe minValue == maxValue is not indentend in the first place and we should fix this problem earlier?

Fix looks ok to me, avoiding the crash is the priority.

PR #405, opened 2024-05-26
GuiTextBox(): Fix buffer overrun from cursor index

segcore:
Fix a buffer overrun from using the global variable textBoxCursorIndex in GuiTextBox().

To reproduce overrun, have two GuiTextBox()es with different buffer sizes, and then the cursor index is shared between them, reading beyond the end of the array of the smaller buffer.

(Found because I run my debug builds with -fsanitize=address and this crashed my program :p )


raysan5:
@segcore Good catch! Thanks for the review!

PR #406, opened 2024-05-27
Create Rollup Box GUI element

Spacerulerwill:
I propose the addition of a `GuiRollupBox` function, which is similar to the `GuiDropdownBox` function but the elements appear above instead of below. This is useful in cases where there is a bottom-level toolbar that still need some kind of select menu, as shown below. With a `GuiDropdownBox` the elements would have fallen off the bottom of the screen.
rollupbar
To implement this, I have refactored the logic of `GuiDropdownBox` into a new function called `GuiDropdownBoxEx,` which accepts an additional parameter for direction. This allows both `GuiDropdownBox` and `GuiRollupBox` to share the same core functionality, only differing by which direction they pass to `GuiDropdownBoxEx`. This is my first proper open source contribution so I apologize in advance for any mistakes.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -619,6 +619,11 @@ typedef enum {
     DROPDOWN_ITEMS_SPACING      // DropdownBox items separation
 } GuiDropdownBoxProperty;
 
+typedef enum {
+    DROPDOWN_DOWN,
+    DROPDOWN_UP
+} GuiDropdownBoxDirection;
+
 // TextBox/TextBoxMulti/ValueBox/Spinner
 typedef enum {
     TEXT_READONLY = 16,         // TextBox in read-only mode: 0-text editable, 1-text no-editable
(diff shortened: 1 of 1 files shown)


raysan5:
@Spacerulerwill This property could be added to `GuiDropdownBox()` but the PR should be redesigned to follow raygui design.

A new property can be added in `GuiDropdownBoxProperty`, for example `DROPDOWN_ROLL_UP`, with default/initial value as 0. Then consider that property directly on `GuiDropdownBox()` with no need to add new controls.

So users will write code as:
```c
GuiSetStyle(DROPDOWNBOX, DROPDOWN_ROLL_UP, 1); // Setup dropdown box to roll-up
GuiDropdownBox(...);
GuiSetStyle(DROPDOWNBOX, DROPDOWN_ROLL_UP, 0); // Reset dropdown box to roll-down
```

raysan5:
@Spacerulerwill Are you reviewing this PR with proposed changes?

Spacerulerwill:
> @Spacerulerwill Are you reviewing this PR with proposed changes?

Apologise for the delay, I forgot about your response. I will review it in the coming days and make the changes

raysan5:
@Spacerulerwill Just added it, I needed it for one of my tools. Thanks anyway!

PR #408, opened 2024-05-30
Fix icon padding adjustment in `GetTextWidth`

kolunmi:
This fixes some issues, especially when using `GuiIconText` with `GuiGroupBox`. Thank you!


raysan5:
@kolunmi Thanks for the review!

PR #410, opened 2024-06-25
Allow passing icons in GuiTabBar

PasVegan:
Currently the icon is set to ICON_FILETYPE_IMAGE for every tab, I don't know why.
But I think it's better for the user to set his own icons


raysan5:
@PasVegan Actually there is no need to add new function parameters, icons can be added by user on text string for every tab.

PR #411, opened 2024-07-05
Align text in the basic controls section

ninadsachania:


raysan5:
@ninadsachania Nice! Thanks for the review!

PR #413, opened 2024-07-11
Support keypad Enter key with GuiValueBox and GuiValueBoxFloat.

mdavisprog:
Wanted to add a little convenience to these functions to allow for the keypad enter key to be used for confirming.


raysan5:
@mdavisprog thanks for the improvement!

PR #418, opened 2024-07-24
Fix modules with compile errors when in C++ mode

foxblock:
Fixed MSVC error messages
"C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax" i.e. (Vector2) -> RAYGUI_CLITERAL(Vector2)
"C4838: conversion from 'int' to 'float' requires a narrowing conversion" i.e. added explicit (float) cast

I have only touched two header files, which one might use as-is in other projects (as I did). I have not changed the actual examples, since they are plain-C and should compile fine anyway.

Cheers.


foxblock:
@raysan5 can this please be merged?

raysan5:
@foxblock Those examples are intended to be used from C code, I'm afraid I'm not changing them for a specific use case.

PR #419, opened 2024-07-25
README: Added move in building for linux

Brian-ED:
Added the move since it restores the state of the files to normal again. I considered copying, decided not to since copying adds an unnecessary file.


raysan5:
@Brian-ED thanks for the improvement!

PR #420, opened 2024-07-28
GuiTextBox(): Do not get previous code point when at start of string

segcore:
Prevents going out of bounds before the beginning of the string


raysan5:
@segcore Thanks! Nice improvement!

PR #430, opened 2024-09-11
Changed the way tooltips work, for better usability

AndreaBoroni:
I felt like there where a couple of issues with the way the Tooltip functionality was implemented:
1. The tooltip is drawn immediately after the control and, in a tight panel of controls, this space can be later occupied by another control, thus hiding the tooltip
2. Every time the tooltps string were added with `GuiSetTooltip(my_tooltip_string)` before a control they had to be reset with another call to `GuiSetTooltip(NULL)` with `NULL` as argument.

This change aims at fixing both.

Point 1:
Now during a frame the `guiTooltipDrawPtr` and `guiTooltipDrawRec` are modified in case a tooltip should activate and at the end of the frame (or whenever the user wants) the tooltip can be drawn by calling `GuiTooltip()`, now taking no arguments and being exposed to the users.

Point 2:
The `guiTooltipPtr` is reset at the end of the controls that allow it. This eliminates te verbosity of calling `GuiSetTooltip(NULL)` after every control.

I tried to be consistent with the naming and the comments, let me know if anything else should change. This will be a breaking change for tooltips, there are no examples using them so maybe we could add one as well. Let me know. I will definitely use my implementation in my projects.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -1380,8 +1380,10 @@ static unsigned int guiIconScale = 1;           // Gui icon default scale (if ic
 
 static bool guiTooltip = false;                 // Tooltip enabled/disabled
 static const char *guiTooltipPtr = NULL;        // Tooltip string pointer (string provided by user)
+static const char *guiTooltipDrawPtr = NULL;    // Tooltip string to be drawn this frame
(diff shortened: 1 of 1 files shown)


raysan5:
@AndreaBoroni I'm afraid this redesign is a breaking change and it would break all the codebases using previous implementation, including all my tools. I'm afraid I'm not merging it.

PR #431, opened 2024-09-12
add ctrl backspace support for deleting words

shakkar23:
The lack of cursor movement and shortcut support is a bit lacking, if you have any requirements for me to add other ctrl support I would be happy to do so!


raysan5:
@shakkar23 Thanks for the addition!

raysan5:
@shakkar23 It would be nice if it could also move from one word to next/previous using CTRL+Cursors, not only used for delete...

Also noticed that when reaching first word of the sentence, it does not delete the first character.

PR #437, opened 2024-10-31
Prevent crash when passing NULL to GuiToggleSlider

planetis-m:
Later in the function there's a if text != NULL check that draws an area only if there's text passed to the function. This prevents crashing otherwise.


raysan5:
@planetis-m thanks for the review!

PR #438, opened 2024-11-02
Added guiAlpha getter

planetis-m:


raysan5:
@planetis-m I think there is no need for this function, alpha is managed on user-side

PR #450, opened 2025-01-28
Fix GuiTextSplit overrunning on max split count

segcore:
When splitting text by delimiters, the index could overrun the result buffer if exceeding the maximum number of splits (default 128). Changed a > to a >= :D

As this function is used internally by various other components, it affects various components.

This did cause a real crash for me (albeit in testing)


raysan5:
@segcore Thanks for the review!

PR #452, opened 2025-02-01
Added drawing of Slider knob in disabled state

ssoher:
The knob rectangle of the the `GuiSliderPro` is not being drawn in disabled state, producing an empty slider. But it makes sense to see the current value of the slider even when it's disabled. So, this PR adds it.


raysan5:
@ssoher Ok, I can see the use cases. Thanks for the review.

PR #454, opened 2025-02-03
Fix mismatch between `if defined` and `define` for `RAYGUI_LINE_MARGIN_TEXT`.

drasmart:
### Changes

- Make the check match the defined macro


raysan5:
@drasmart Good catch! Thanks for the fix!

PR #457, opened 2025-02-05
Fixed incompatible-pointer-types errors in examples

Developer-Harshit:
Checkout this for more info
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wincompatible-pointer-types


raysan5:
@Developer-Harshit Thanks but I'm addressing those issue with another approach, just removing the const from the strings array definition.

Issue #399, opened 2024-04-23
Need to fix unicode input on GuiTextBox


electrolys:
gui textbox doesn't handle adding and removing characters properly and causes an underflow (at adding) and overflow (at removal)

at lines 
https://github.com/raysan5/raygui/blob/6f532337ff2d6dc14ba3a5f5b7ee63f9d90ff13b/src/raygui.h#L2582
https://github.com/raysan5/raygui/blob/6f532337ff2d6dc14ba3a5f5b7ee63f9d90ff13b/src/raygui.h#L2562C21-L2562C113

you are just comparing the variable i with the text length or cursor when you really need to >= cursor+codepointsize and <= length-codepointsize

raysan5:
@electrolys Good catch! Feel free to send a PR to fix it!

raysan5:
It seems there is no interest in this fix...

PR #458, opened 2025-02-06
Do not overflow when deleting Unicode in GuiTextBox()

SirNate0:
https://github.com/raysan5/raygui/issues/399

I have not tested it (I don't use Unicode input myself), I just saw the issue and thought it should be a simple fix.

I also noted that there are some inconsistencies in whether `autoCursorCooldownCounter` is tested `>` or `>=` `RAYGUI_TEXTBOX_AUTO_CURSOR_COOLDOWN`. Were these deliberate, or should they be changed as well? And was it intended that the re-triggering on hold would not happen until `2*RAYGUI_TEXTBOX_AUTO_CURSOR_COOLDOWN` (because of the `autoCursorDelayCounter++` at the first line of the if, the `%RAYGUI_TEXTBOX_AUTO_CURSOR_COOLDOWN` will only be zero when it is `2*` and not `1*`)?

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -2608,11 +2608,11 @@ int GuiTextBox(Rectangle bounds, char *text, int textSize, bool editMode)
                 {
                     int nextCodepointSize = 0;
                     GetCodepointNext(text + textBoxCursorIndex, &nextCodepointSize);
-
-                    // Move backward text from cursor position
+                    
+                    // Decrease length and move text back to current cursor position
+                    textLength -= nextCodepointSize;
                     for (int i = textBoxCursorIndex; i < textLength; i++) text[i] = text[i + nextCodepointSize];
 
-                    textLength -= codepointSize;
                     if (textBoxCursorIndex > textLength) textBoxCursorIndex = textLength;
 
                     // Make sure text last character is EOL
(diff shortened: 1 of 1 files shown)


foxblock (changes requested):
I just noticed this PR overlaps with [mine](https://github.com/raysan5/raygui/pull/462), which was already merged🙈
This is probably where the merge conflicts come from. Sorry about that (I must have overlooked your PR).

I tested the changes you made

- changing codepointSize to prev/nextCodepointSize is correct and what my PR changed as well
- when handling BACKSPACE moving the decrement by prevCodepointSize before the loop (which moves the data), is also correct (I missed that one)

The changes related to BACKSPACE+CTRL are fine in theory, but the whole section is not working correctly anyway (it will never delete the first char in a string, it does not stop at punctuation - which it does on Windows in other applications, it calls isspace potentially with UTF-8 control chars which is undefined behavior and raise an assertion in debug mode for me)
So I would like to attempt to completely rewrite that block it in the coming days.

Should I just make a new PR incorporating your changes as well or can I somehow add changes to this PR?

raysan5:
@SirNate0 @foxblock I'm afraid I just missed this PR, sorry for the inconvenience... I'm closing this PR. Please, open a new one with the required imrprovement, I'll try to review it asap.

PR #462, opened 2025-02-25
Fix GuiTextBox handling of codepoints when deleting text and clicking with mouse

foxblock:
The codeblocks handling the DELETE and BACKSPACE keys used the wrong codepointSize variable (the one for the key being pressed, instead of the codepoint being deleted).

Clicking with mouse iterated through the array without respect to the codepoint size.

This PR fixes both.


raysan5:
@foxblock thanks for the fix! 😄

PR #463, opened 2025-02-25
GuiTextBox paste support

foxblock:
Handling paste similarly to how standard input boxes on Windows behave:

- will try to insert as much of the clipboard text as possible at the given cursor position, without overflowing the buffer
- will only add full codepoints
- will stop at any control character in the input (e.g. will only insert until a newline character in encountered, unless multiline is true - to prepare for future multiline support)

Assumes GetClipboardText() always returns either NULL or an UTF-8 encoded string (which sadly is not clearly documented, but seems to be true for at least the GLFW and SDL backends).

Tested with various extended latin characters and emojis, which all work as expected.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -2594,9 +2594,6 @@ int GuiTextBox(Rectangle bounds, char *text, int textSize, bool editMode)
                     }
                     if (pasteLength > 0)
                     {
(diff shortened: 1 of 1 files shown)


raysan5:
@foxblock Nice improvement! Thanks!

PR #464, opened 2025-02-26
Fix GuiSlider not drawing handle

foxblock:
Here is the change that caused this:
https://github.com/raysan5/raygui/commit/46fd886f18c6bed1017f1cd6e9db873b2816a828#diff-d01adeecbcae2a1e13b6832e327d5f84b9ec0e13aff89630006e99a24a9f2273L4304

I believe this change was accidental. 
At least I could not find an instance where the VALUEBOX, SLIDER_WIDTH style was accessed.


raysan5:
> I believe this change was accidental.

@foxblock It seems so! Thanks for the review!

PR #465, opened 2025-02-28
fixed border and padding offset issue caused by BORDER_WIDTH

pmHaugen:
I've tried to implement the fix according to how I think it is meant to work, so I hope my preception is accurate.
For example:
image
The border around the filled bar is increased with the padding amount so it can reach its max value at the same time as the progress bar as it is squshed together from all direction now.

**Issue I tried to fix** 
- The rectangles on the right side of the GuiProgressBar() becomes weirdly offset by adding BOREDER_WIDTH
- The Fill bar gets padded incorrectly.

image

**Potential problems** 
- Setting padding or boarder width higher than the size of the GuiProgressBar() it behaves incorrecly. Protecting against this just hides the user error imo
- I have not tested that the boarder rectangles are not overlapping and I had a feeling from watching the previous code that they were set up to overlap in the corners, I don't know if this is wastefull or causes any problem...

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -3308,16 +3308,16 @@ int GuiProgressBar(Rectangle bounds, const char *textLeft, const char *textRight
     if (value == NULL) value = &temp;
 
     // Progress bar
-    Rectangle progress = { bounds.x + GuiGetStyle(PROGRESSBAR, BORDER_WIDTH),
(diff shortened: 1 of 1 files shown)


raysan5:
@pmHaugen thanks for the improvement!

raysan5:
@pmHaugen Just notices some missalignment after merging this PR, please, could you review it?

image

pmHaugen:
No problem, I can revert it to ignore left and right padding again. I guess optimaly there would be an option for this as atleast for me it looked more "correct" this way but I am no designer :) I'll look into as soon as I can

PR #466, opened 2025-02-28
GuiTextBox() rework

foxblock:
Sooo, this PR became a bit bigger than I initially planned. Sorry about that.
I tried to keep the individual changes each in their own commit, so if the sum of changes is a bit too much to review, maybe you can go commit by commit.
Let me know, if I should do multiple PRs for something like this in the future.

Anyway, this does the following:

- Implement the correct bounds checking by @SirNate0
- Fix multiple problems related to CTRL+BACKSPACE handling. The behaviour should now mostly match nodepad, Notepad++ or other Windows applications with "standard" text processing (the one exception being holding CTRL+BACKSPACE)
- Implement that behaviour also for CTRL+DELETE, CTRL+LEFT and CTRL+RIGHT
- Simplify handling of "auto cursor" (reduce code a bit) and bring that behaviour also more in-line with how Notepad++ and others feel in that regard

Doing this, I noticed that VSCode has implemented this differently from the mentioned applications. Probably because they went the naive route and implemented the left-looking and right-looking variants the same (you will see what I mean, when you look at the code)... sigh.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -2629,8 +2629,42 @@ int GuiTextBox(Rectangle bounds, char *text, int textSize, bool editMode)
             // Move cursor to end
             if ((textLength > textBoxCursorIndex) && IsKeyPressed(KEY_END)) textBoxCursorIndex = textLength;
 
+            // Delete related codepoints from text, after current cursor position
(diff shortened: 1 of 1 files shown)


raysan5:
@foxblock Wow! A big set of changes! Looks good! Thanks for the improvements!

Issue #468, opened 2025-02-28
[potrable_window] Window may briefly flicker, if user moved it via different means than dragging the bar


sleeptightAnsiC:
Portable Window Example does not expect that someone can move a window in different way than dragging the bar. In this case the window will briefly flicker, if user then tries to move it by dragging. I guess, this probably never occurs on Windows where Desktop is very limited, but on Linux it's a different story.

https://github.com/user-attachments/assets/65e08348-bf51-4ef9-b8cb-1a6fd289a772

I was able to simply fix this with following patch:

```patch
diff --git a/examples/portable_window/portable_window.c b/examples/portable_window/portable_window.c                                                                                                                              
index 8194e64..a7d3c9a 100644                                                                                                                                                                                                     
--- a/examples/portable_window/portable_window.c                                                                                                                                                                                  
+++ b/examples/portable_window/portable_window.c                                                                                                                                                                                  
@@ -57,6 +57,7 @@ int main()                                                                                                                                                                                                      
         {                                                                                                                                                                                                                        
             if (CheckCollisionPointRec(mousePosition, (Rectangle){ 0, 0, screenWidth, 20 }))                                                                                                                                     
             {                                                                                                                                                                                                                    
+                windowPosition = GetWindowPosition();                                                                                                                                                                            
                 dragWindow = true;                                                                                                                                                                                               
                 panOffset = mousePosition;                                                                                                                                                                                       
             }
```
Here's how it should look like:

https://github.com/user-attachments/assets/a117ce2c-7d10-4f9a-a55d-33246b1ff5ee

(I'll create PR for this in a moment)

PR #469, opened 2025-02-28
[portable_window] fix window flicker when moving it without dragging the bar.

sleeptightAnsiC:
Fixes: https://github.com/raysan5/raygui/issues/468


raysan5:
@sleeptightAnsiC thanks for the review!

PR #470, opened 2025-03-06
Export GuiSliderPro and add static to GetTextLines (for shared libs)

pfeodrippe:
I was having issues using it in conjunction with raylib (in Windows), so added these annotations here.

Let me know if it's okay, thanks for the project, Ray!!


raysan5:
@pfeodrippe Merging the change but not sure if I want to expose the Pro version of GuiSlider(), maybe revert it in the future.

pfeodrippe:
> @pfeodrippe Merging the change but not sure if I want to expose the Pro version of GuiSlider(), maybe revert it in the future.

For sure o/ Whatever you prefer :D thanks, Ray!

PR #471, opened 2025-03-09
Fix building and cleaning examples on OpenBSD.

voutilad:
On OpenBSD, the X11 libraries are under `/usr/X11R6/lib`.

For the `clean` target, `find(1)` can be used similar to the OSX platform logic, but with using octal-based permissions.


raysan5:
@voutilad thanks for the review!

PR #473, opened 2025-03-14
removed left-right padding from GuiProgressBar

pmHaugen:
As requested by @raysan5 I removed the left-right padding I wrongfully added in my last pull request

Thanks for reaching out so I could fix the issue : )

image


raysan5:
@pmHaugen thanks for the quick review and sorry for the invonvenience!

Issue #476, opened 2025-03-14
Possible integer overflow in icon loading


AmityWilder:
## Description

The following arithmetic can, depending on the C compiler version, cause an integer overflow for icons larger than 181x181.

https://github.com/raysan5/raygui/blob/1536ae35c7b42d863135f4181fd2a225e531f68b/src/raygui.h#L4564

https://github.com/raysan5/raygui/blob/1536ae35c7b42d863135f4181fd2a225e531f68b/src/raygui.h#L4614

```c
/* ... */ iconCount*(iconSize*iconSize/32) /* ... */
```

`iconSize` is a (signed) short -- max positive value: 32767
```math
\begin{align*}
\sqrt{32767} &\approx 181.01 \\
182*182 &= 33124 \\
&\equiv \mathbf{-32412} \pmod{\pm 32767}
\end{align*}
```

## Proposal

In both instances of this problem, the resulting value is being passed to an integer type wide enough to accept values greater than 181x181. This cast should occur prior to multiplying the `short`s together, so that they don't overflow before the implicit cast.

```c
/* ... */ iconCount*((int)iconSize*(int)iconSize/32) /* ... */
```

PR #477, opened 2025-03-14
Fix integer overflow potential

AmityWilder:
Related to #476


raysan5:
@AmityWilder Thanks for the review. Note that I try to avoid `size_t` usage, I'm replacing it.

Issue #478, opened 2025-03-14
Potential memory leak/stack free in `GuiLoadStyleDefault()`


AmityWilder:
## Description

The function `GuiLoadStyleDefault()` uses `RL_FREE` instead of `RAYGUI_FREE`, creating the possibility for `RAYGUI_MALLOC`/`RAYGUI_CALLOC` to be used for allocating memory in a way that is incompatible with `RL_FREE`.

https://github.com/raysan5/raygui/blob/1536ae35c7b42d863135f4181fd2a225e531f68b/src/raygui.h#L4446-L4447

This has the potential to cause either:

- A memory leak, if `RAYGUI_MALLOC` returns `malloc()`'d memory that goes un-`free`'d by a custom implementation of `RL_FREE` that deliberately doesn't deallocate memory, expecting it to be a stack pointer.
- A stack-free, if `RL_FREE` calls `free()` on a custom `RAYGUI_MALLOC` implementation that returns a pointer to the stack.
- No actual problem--but some headache--if `RAYGUI_MALLOC` and `RL_FREE` are compatible, but `RAYGUI_MALLOC` is customized to have a breakpoint or message for monitoring allocations while `RL_FREE` doesn't (or vice-versa).

AmityWilder:
## Addendum

Perhaps this might be evidence that #472 should be applied to Raylib core as well, to encourage libraries using raylib to perform an audit of these cases where they ought to be using `MemAlloc`/`MemFree`, or mistakenly use `RL_`-memory functions when they provide their own?

The drawback to this is that it might break compatibility for those libraries, if their maintainers don't end up performing such an audit.

PR #479, opened 2025-03-14
Replace `RL_FREE` with `RAYGUI_FREE`

AmityWilder:
Related to #478


raysan5:
@AmityWilder Thanks for the review

PR #480, opened 2025-03-15
Making GetTextWidth part of the public API

guifes:
GetTextWidth is useuful to be called publicly.
E.g: when you want to get the final horizontal size of a Slider that has textBefore and textAfter.


raysan5:
@guifes Ok, I'm exposing the function

PR #485, opened 2025-04-20
Add minus toggle for GuiValueBox & GuiValueBoxFloat

nezvers:
It was impossible to enter negative number.
Now pressing KEY_MINUS toggles sign for the value when value is not 0. Due to original implementation it's not possible to start entry with minus symbol or have added minus while the value is 0.


raysan5:
@nezvers Nice improvement! Thanks!

PR #497, opened 2025-07-13
Removed unpaired multi-line comment end

Brian-ED:
An unpaired multi-line comment end found it's way infront of a comment. I simply removed it.


raysan5:
@Brian-ED Good catch! Thanks for the review!

PR #500, opened 2025-07-28
Include of raygui_icons.h

wwderw:
On a couple of examples, although commented out, it was looking for "gui_icons.h" versus "raygui_icons.h".  I only saw two examples of this.  This PR is to fix that typo.


raysan5:
@wwderw Actually those examples are intended to illustrate how to include custom gui icons, not the raygui icons provided, so I prefer to keep the name of the file not related to raygui icons.

Issue #514, opened 2025-10-25
Null pointer derefrence in GuiLoadStyleDefault function


lgdt-mem-takeover:
Hey, apologies if there is some template for reporting that I am not using ( haven't seen an obvious link to one ). The GuiLoadStyleDefault segfaulted on me when  using my custom allocator, and I noticed that the function uses RAYGUI_FREE() ( which uses free() from stdlib I assume). 

```  
if (guiFont.texture.id != GetFontDefault().texture.id)
    {
        // Unload previous font texture
        UnloadTexture(guiFont.texture);
	RAYGUI_FREE(guiFont.recs);
	RAYGUI_FREE(guiFont.glyphs);
        guiFont.recs = NULL;
        guiFont.glyphs = NULL;

        // Setup default raylib font
        guiFont = GetFontDefault();

        // NOTE: Default raylib font character 95 is a white square
        Rectangle whiteChar = guiFont.recs[95];

        // NOTE: We set up a 1px padding on char rectangle to avoid pixel bleeding on MSAA filtering
        SetShapesTexture(guiFont.texture, RAYGUI_CLITERAL(Rectangle){ whiteChar.x + 1, whiteChar.y + 1, whiteChar.width - 2, whiteChar.height - 2 });
    }
```

I am not sure how free() handles it, but if a pointer is NULL already, you shouldn't be allowed to free it again. My allocator does store some metadata at the base of the pointer, so when I try to access it, that will obviously also not work, if it's NULL. Regardless, this is basically a double free so I just guarded it.

```

    if (guiFont.texture.id != GetFontDefault().texture.id)
    {
        // Unload previous font texture
        UnloadTexture(guiFont.texture);
	if (guiFont.recs != NULL) RAYGUI_FREE(guiFont.recs);
	if (guiFont.glyphs != NULL) RAYGUI_FREE(guiFont.glyphs);
        guiFont.recs = NULL;
        guiFont.glyphs = NULL;

        // Setup default raylib font
        guiFont = GetFontDefault();

        // NOTE: Default raylib font character 95 is a white square
        Rectangle whiteChar = guiFont.recs[95];

        // NOTE: We set up a 1px padding on char rectangle to avoid pixel bleeding on MSAA filtering
        SetShapesTexture(guiFont.texture, RAYGUI_CLITERAL(Rectangle){ whiteChar.x + 1, whiteChar.y + 1, whiteChar.width - 2, whiteChar.height - 2 });
    }
```

[...]

PR #518, opened 2025-12-15
Fix issue #514: prevent crash with custom allocators in GuiLoadStyleDefault

grerfou:
### Description

This PR fixes issue #514 where GuiLoadStyleDefault could crash when using a custom allocator.  
The problem occurs because RAYGUI_FREE was called on guiFont.recs and guiFont.glyphs without checking if the pointers were NULL.

### Changes

- Added checks before freeing pointers:

if (guiFont.recs != NULL) RAYGUI_FREE(guiFont.recs);
if (guiFont.glyphs != NULL) RAYGUI_FREE(guiFont.glyphs);
guiFont.recs = NULL;
guiFont.glyphs = NULL;

- Added a comment explaining the fix.
- Ensures compatibility with standard malloc/free as well as custom allocators.

### Testing

- Verified that with a custom allocator, the crash no longer occurs.
- Verified normal behavior with standard malloc/free.

This fix is minimal and does not change any other functionality.

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -4509,8 +4509,9 @@ void GuiLoadStyleDefault(void)
     {
         // Unload previous font texture
         UnloadTexture(guiFont.texture);
-        RAYGUI_FREE(guiFont.recs);
-        RAYGUI_FREE(guiFont.glyphs);
+        // Fix issue #514: avoid crashing with custom allocators by checking NULL before free 
+        if (guiFont.recs != NULL) RAYGUI_FREE(guiFont.recs); 
+        if (guiFont.glyphs != NULL) RAYGUI_FREE(guiFont.glyphs); 
         guiFont.recs = NULL;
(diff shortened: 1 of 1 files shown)


grerfou:
Thanks for the clarification, you are absolutely right.

After further investigation, I agree this is not a raygui bug.
The crash happens because the custom allocator used in issue #514
does not handle free(NULL), which violates the C standard contract.

Raygui is behaving correctly by calling RAYGUI_FREE unconditionally.
The proper fix is to update the custom allocator to safely ignore NULL pointers.

This was my first open source contribution, and I misunderstood where the responsibility
should lie. Thanks for the patience and for taking the time to explain it clearly.

I will close this PR accordingly.

grerfou:
@raysan5 I’m a beginner in C and would love to start contributing to raygui. 
Could you suggest a simple issue or area suitable for a first contribution? Thank you!

PR #526, opened 2026-02-10
Fix GuiSlider jumping to mouse position on click

### LINKED ISSUE #474: `GuiSlider()` when clicking on the slider-bar for draging, position changes to become mouse-centered
It would be nice if it could be clicked to start draging from clicked position, with no bar repositioning.

grerfou:
Fixes #474 

## Problem
When clicking on the slider track (not on the handle), the slider would immediately jump to the mouse position before dragging could begin.

## Solution
- Store the initial offset between mouse position and slider position when clicking
- Apply this offset during dragging to prevent the jump
- The slider now drags smoothly from its current position without repositioning

## Testing
Created a test program that demonstrates the fix works for both `GuiSlider()` and `GuiSliderBar()`.

**Before:** Clicking anywhere on track caused slider to jump to that position  
**After:** Clicking starts dragging from current slider position, no jump occurs

**Note:** Opened as draft pending confirmation from @raysan5  on the expected behavior (see issue discussion). Will mark as ready for review once confirmed.


raysan5:
I'm afraid this issue has no solution other than adding a static variable to track position, I prefer to avoid that route.

PR #529, opened 2026-03-04
Don't double define _CRT_SECURE_NO_WARNINGS

JeffM2501:
This PR ensures that _CRT_SECURE_NO_WARNINGS is not defined again if it is already defined.
This mostly suppresses warnings.


raysan5:
@JeffM2501 thanks for the review!

PR #537, opened 2026-04-08
Fix wrongly casting a parameter as `const char**` when calling `GuiListViewEx`

charlyalizadeh:
Fix wrongly casting a parameter as `const char**` when calling `GuiListViewEx` (the function takes a `char**`, no const needed).


raysan5:
@charlyalizadeh thanks for the review

PR #539, opened 2026-04-11
Suppress MSVC C4996 warnings in implementation section

shyraptor:
When raygui.h is included after raylib.h (the common case), MSVC emits
C4996 warnings on every fopen/sscanf call inside the implementation section:

    raygui.h(4376): warning C4996: 'fopen': This function or variable may be unsafe.
    raygui.h(4397): warning C4996: 'sscanf': This function or variable may be unsafe.
    raygui.h(4408): warning C4996: 'sscanf': This function or variable may be unsafe.

raygui already defines _CRT_SECURE_NO_WARNINGS (line 361) to prevent this,
but it has no effect when stdio.h was already pulled in by an earlier header.
MSVC's CRT decides at first include whether to mark these functions deprecated;
a later define cannot undo that.

This PR adds a #pragma warning(push/disable/pop) block around the
RAYGUI_IMPLEMENTATION section, which suppresses C4996 at the call site
regardless of include order. The suppression is scoped — it does not
leak into user code.

Tested on MSVC 19.44 (Visual Studio 2022 Build Tools).

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -1096,6 +1096,11 @@ typedef enum {
 
 #if defined(RAYGUI_IMPLEMENTATION)
 
+#if defined(_MSC_VER)
+    #pragma warning(push)
+    #pragma warning(disable: 4996)  // Disable MSVC C4996 warning for standard C functions (fopen, sscanf, etc.)
+#endif
+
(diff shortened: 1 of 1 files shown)


raysan5:
@shyraptor I can't reproduce, I use `raygui.h` with all my tools, with VS2022, including it after `raylib.h` and I can't get those warnings.

PR #540, opened 2026-04-12
`GuiGetTextWidth()`, fix to measure multi-line text correctly

### LINKED ISSUE #415: GetTextWidth returns incorrect value for multi-line strings
`GetTextWidth` only calculates the length of the first line instead of all lines. This also affects `GuiMessageBox` (and possibly other GUI functions that I haven't gotten around to testing).

Example 1 with `GetTextWidth`
```c
int main(void)
{
	SetConfigFlags(FLAG_VSYNC_HINT);
	InitWindow(800, 450, "Minimal Reproducible Example -- GetTextWidth");

	char exampleText[] = "Line 1\nL i n e  2";

	while (!WindowShouldClose())
	{
		// ===== DRAW ====
		BeginDrawing();
		ClearBackground(WHITE);

		// ===== GUI =====

		int width = GetTextWidth(exampleText);

		// This would ideally draw a box around the example text, but instead the text gets cut off
		GuiDrawRectangle((Rectangle) {20, 20, width + 6, 30}, 2, BLACK, WHITE);
		GuiLabel((Rectangle) {22, 20, width, 30}, exampleText);

		// ===== END GUI ====

		EndDrawing();
		// ===== END DRAW ====
	}

	CloseWindow();
}
```

Example 2 with `GuiMessageBox`
```c
int main(void)
{
	SetConfigFlags(FLAG_VSYNC_HINT);
	InitWindow(800, 450, "Minimal Reproducible Example -- GuiMessageBox");

	char exampleText[] = "Line 1\nL i n e  2";

	while (!WindowShouldClose())
	{
		// ===== DRAW ====
		BeginDrawing();
		ClearBackground(WHITE);

		// ===== GUI =====

		// This would ideally draw the example text in the message box, but it instead gets
		// cut off even though there is plenty of space
		GuiMessageBox((Rectangle) {20, 20, 250, 100}, "#191# Error!", exampleText, "OK");

		// ===== END GUI =====

		EndDrawing();
		// ===== END DRAW =====
	}

	CloseWindow();
}
```

shyraptor:
Resolves #415.

`GuiGetTextWidth()` was stopping at the first `'\n'`, so auto-sized widgets (`GuiLabel`, `GuiMessageBox`, ...) clipped multi-line text.

## Fix

- New static `GetLineWidth()` helper — single-line measurement, identical to the old `GuiGetTextWidth()` body except for the icon corrections below.
- `GuiGetTextWidth()` becomes a wrapper that iterates `'\n'`-separated lines and returns `max(GetLineWidth(line))`.
- `GuiDrawText()` now calls `GetLineWidth()` directly for per-line text positioning, preserving its existing alignment behavior byte-for-byte (the icon is already stripped by `GetTextIcon()` before measurement, so `GetLineWidth()` takes no icon path there).

## Incidental corrections in `GetLineWidth()`

Both corrections bring the icon-marker handling in line with the authoritative `GetTextIcon()`:

1. **Off-by-one in icon skip** — `text += (pos + 1)` instead of `text += textIconOffset`. The old code set `textIconOffset` to the *index of the closing `#`* and then advanced by that, leaving the closing `#` in the measured string. This over-measured every icon-prefixed label by ~1 glyph.
2. **Digit-only icon detection** — the marker is recognised only as `#` + 1..3 digits + `#`. The old detector accepted any character between the hashes, so e.g. `"#abc#Text"` was misclassified as an icon and the label width was computed incorrectly.

## MRE (from #415)

Both examples in the issue body now render with the border/message box fully enclosing the multi-line text.


raysan5:
@shyraptor thanks for the review!

PR #543, opened 2026-04-15
Export symbols on *nix when building shared libraries

DhruvaSambrani:
Like raylib, when raygui is built with `BUILD_LIBTYPE_SHARED`, the `visibility` of symbols tagged with `RAYGUIAPI` need to be set to `default`. This patch does that.


raysan5:
@DhruvaSambrani thanks for the review

PR #544, opened 2026-04-27
Fix minimum vertical scroll value in GuiScrollPanel()

### LINKED ISSUE #488: Different minimum value of x and y in viewscroll of GuiScrollPanel.
I am trying to use the GuiScrollPanel.
If I move the horizontal and vertical scroll forward and then backward, the minimum value of viewscroll vector's x and y are different, 1 and 0 respectively.  I checked it with the below code that prints it while moving the scroll .
Does this have intentional meaning or a bug?

```
Rectangle view = { 0 };
GuiScrollPanel(Rectangle{ 560, 25, 102, 354 }, NULL, Rectangle{ 560, 25, 300, 1200 }, & viewScroll, & view);

printf("scroll.x : %f \n", viewScroll.x);
printf("scroll.y : %f \n", viewScroll.y);
```

js-lm:
Fix #488

Fixed the issue in GuiScrollPanel() where the horizontal and vertical clamped to different minimum values when scrolled fully back.

[before.webm](https://github.com/user-attachments/assets/ac046a1f-aec8-49c6-8e52-f31b268baf24)

[after.webm](https://github.com/user-attachments/assets/8e1924a6-68ee-440f-8e0a-a90a8f9150c2)


raysan5:
@js-lm thanks for the review

Issue #402, opened 2024-05-07
Define a consistent convention on controls return values


raysan5:
raygui was redesigned on v4.0 to always return an `int` as "control result value" while keeping the control required variables as function parameters, usually passed by reference.

Despite the redesign, there was no rule defining/unifying that "control result value" that should be returned by every control. Here some possible ideas (ref: #371):

```c
typedef enum GuiControlResult {
    RESULT_NONE = 0,
    RESULT_FOCUSED = 1,
    RESULT_CLICKED = 2,
    RESULT_SCROLLED = 3,
    RESULT_CUSTOM01 = 4,
    RESULT_CUSTOM02 = 5,
} GuiControlResult;
```

Every control has to be carefully analized to see what are the possible internal states that could be returned. It would be nice to unify the possible return values but also allow per-control custom return values.

Note that some controls already return specific results, reviewing this functionality could be a breaking change:

 - `GuiButton()`: Returns `1` when clicked --> `RESULT_CLICKED`?
 - `GuiTextBox()`: Returns `true` when edited, used to change `editMode`
 - ...

ohmtal:
maybe
```
typedef enum GuiControlResult {
    RESULT_NONE = 0,
    RESULT_CHANGED = 1, // << does not break current behavior
    RESULT_CLICKED = 2,
    RESULT_KEYDOWN = 3, // using focus  
    RESULT_FOCUSED = 4,
    RESULT_SCROLLED = 5,
    RESULT_CUSTOM01 = 6,
    RESULT_CUSTOM02 = 7,
} GuiControlResult;
```

raysan5:
Implemented on https://github.com/raysan5/raygui/commit/14cb4c4c0581eaa8d0f546a2b9c6b1e84b085449, kept it simple for now. Improving it requires a redesign to all controls that expect `bool editMode`, to be passed by reference, left for a future review...

PR #545, opened 2026-05-06
GuiComboBox did not set result on value changed

ohmtal:
Not much to say. added result = 1; when clicked.


raysan5:
@ohmtal convention to follow has not been defined yet: https://github.com/raysan5/raygui/issues/402

Issue #546, opened 2026-05-07
[FEATURE] Include modern .rgs file for default style


jgabaut:
## Request

Could a modern `.rgs` binary file for default style be added to the repo?

Currently, it's the only one missing.
I did some digging around and was unable to produce it myself using different versions of `rGuiStyler` with the older files from the mentioned commits.

I saw this commit reviewed all styles to support `ISO-8859-15` charset (as much as possible), so maybe it was impossible to produce a `.rgs` at that time?
- https://github.com/raysan5/raygui/commit/db466282f811771c355655d8f81ad27ca1011548

## History

| Description| Raygui version      | Commit      |
| ------------- | ------------- | ------------- |
| Added `v3.1` text file `default.rgs` | `2.6` | https://github.com/raysan5/raygui/commit/5b08be808fad7ad43ffb05d754f2f41a88f1e9c7 |
| Updated `v3.1` text file `default.rgs` properties | `2.6` | https://github.com/raysan5/raygui/commit/1dd5f423b8eb96baec0c5f1c555e1859c8423707 |
| Add binary `.rgs`, text `.h` files (`v3.1`) for all styles BUT default. Default only gets the rename of old text file `default.rgs` to `default.txt.rgs`  | `3.1` | https://github.com/raysan5/raygui/commit/23121c5454f2a709e733dad2e2fbbd565b0e410d |
| Review for support of `ISO 8859-15`. Update binary `.rgs`, text `.h` files to `v4.0` and prepend `style_` to file names for all styles BUT default. Default only gets binary file `style_default.old.rgs`, probably `v3.x` | `4.0` | https://github.com/raysan5/raygui/commit/db466282f811771c355655d8f81ad27ca1011548 |
| Drop all `.old.rds` `v3.x` binary files | `4.0` | https://github.com/raysan5/raygui/commit/8d299c3c5016b5e697c7c6b6867bc7ab7d16fafa |

If there is no blocker on this I would really appreciate it being added.
It would allow my tool with theme loading to avoid treating the default style with special cases (like doubling the font size while not doing if for the others), since I could just use a doubled-base-size theme obtained from the default `.rgs`.
See:
- https://github.com/JGGR/F-value/blob/f64415eb92e52851945770f704ce8ef8235e11c9/src/app/controller.rs#L204-L220
- https://github.com/JGGR/F-value/blob/f64415eb92e52851945770f704ce8ef8235e11c9/src/app/core.rs#L475-L483

[...]

PR #547, opened 2026-05-07
chore: add style_default.rgs from rGuiStyler 6.0

jgabaut:
- Tries closing #546 

I don't think this works.

Loading this file in my tests only breaks the font of the current theme, and does nothing else useful.


raysan5:
I think there is no need for this file to be provided, it's always available on raygui with `GuiLoadStyleDefault()`

jgabaut:
> I think there is no need for this file to be provided, it's always available on raygui with `GuiLoadStyleDefault()`

I understand. I have updated the mentioned issue to better convey what I was actually pursuing.

The problem Is perhaps more that I haven't been able to properly export a _slightly_ custom file, not really the fully default style (which one could already obtain from`GuiLoadStyleDefault()`, for the same effect of loading from `.rgs`).

Thanks for your time and sorry for the confusing approach.

Issue #548, opened 2026-05-07
[FEATURE] Expose GuiLoadStyleFromMemory


jgabaut:
## Request

I would like to use the `GuiLoadStyleFromMemory` function.

Currently, the `GuiLoadStyleFromMemory` function is private and not available to usercode.
- https://github.com/raysan5/raygui/blob/6d2b28ff748158be0d63d07988d5d0672905dedf/src/raygui.h#L4848
They must resort to `GuiLoadStyleFromFile` (which, ultimately, still uses `GuiLoadStyleFromMemory`):
- https://github.com/raysan5/raygui/blob/6d2b28ff748158be0d63d07988d5d0672905dedf/src/raygui.h#L4372-L4477

While it seems that the body of `GuiLoadStyleFromMemory` has some hints 
- https://github.com/raysan5/raygui/blob/6d2b28ff748158be0d63d07988d5d0672905dedf/src/raygui.h#L4890-L4892

about its implementation being highly dependant on `raylib`, this does not change the fact that the same code section can be reached publicly with the `FromFile` method.

This limitation ultimately limits the ability to embed styles in programs using `raygui`.

Switching between theme definitions obtained from embedding the style files requires some funny code to:
- Write a temporary file from the already-in-static-memory bytes
- Load it with the file API
- Delete the temp file (or not)

Which is prone to error and arguably just worse than exposing the existing, and ultimately already used (same-control-flow-reachable), `GuiLoadStyleFromMemory`.

## Usecase

I have prepared a draft PR for my usecase, involving the rust bindings.

- https://github.com/JGGR/F-value/pull/21
- The diff is short but telling [a story I can't express in one line here](https://github.com/JGGR/F-value/pull/21/changes#diff-374a6f355fb52edd27af4a0ac4ae5262343802ce01d393498c372ee295618938)
- In the draft, I could also drop my dependency on `uuid`, which is currently needed only for this tempfile-load-trash [chicanery](https://www.youtube.com/watch?v=PuZ34IeY_L0).

## Proposed solutions

I don't really see a real drawback to exposing the function since the same logic is reachable with the `GuiLoadStyleFromFile()` method.
As a side note, it seems that currently the `dataSize` argument is not used.

If the function is deemed "unsafe" for some reason, I would like at least a specific build flag to turn a macro guard on an expose it...

But ultimately, having this API just available, without jumping through hoops (for no better or worse guarantees), would be better.

raysan5:
Done.

[...]

PR #549, opened 2026-05-07
feat: expose GuiLoadStyleFromMemory()

jgabaut:
- Tries closing #548 

Naive PR, just straight up moves the function to near the already public `GuiLoadStyle()`.

Conflates exposition (dropping `static`) with styling (proper placement in source file).

The diff is hard to read, but I basically did just that.

Two different commits could be better, maybe?

I could rebase the PR to have a more readable diff that only removes the `static`, and a second commit that moves declaration and definition (large diff).


raysan5:
@jgabaut I think it can be exposed.

jgabaut:
> @jgabaut I think it can be exposed.

Thank you! I appreciate it, knowing expanding the supported interface is not free.

Issue #550, opened 2026-05-11
Decouple internal widget icon mappings


js-lm:
Currently, when using `RAYGUI_CUSTOM_ICONS`, the custom enum must replicate raygui's internal naming conventions exactly, as the values are hardcoded like

```
result = GuiButton(closeButtonRec, GuiIconText(ICON_CROSS_SMALL, NULL));
```

or

```
if (GuiButton(leftButtonBound, GuiIconText(ICON_ARROW_LEFT_FILL, NULL))) tempValue--;
```

This means if I use a different naming convention or want to use a different icon, I either have to make my enum match the hardcoded values or fork raygui just to change the icons.

It would be nice to add a layer of macro for icons used internally by raygui widgets, something like this

```
#if !defined(GUI_ICON_WINDOW_CLOSE)
    #define GUI_ICON_WINDOW_CLOSE    ICON_CROSS_SMALL
#endif

#if !defined(GUI_ICON_SPINNER_LEFT)
    #define GUI_ICON_SPINNER_LEFT    ICON_ARROW_LEFT_FILL
#endif
```
```
result = GuiButton(closeButtonRec, GuiIconText(GUI_ICON_WINDOW_CLOSE, NULL));
```
```
if (GuiButton(leftButtonBound, GuiIconText(GUI_ICON_SPINNER_LEFT, NULL))) tempValue--;
```

raysan5:
@js-lm there are too many icons, I don't want to pollute code with that kind of checks, sorry

PR #551, opened 2026-05-16
Decouple internal widget icon mappings

js-lm:
Implement #550

Maybe this should be split into two pr, but since the changes are closely related, I decided to keep them in one

The first commit replaced the hardcoded icon ids strings with `GuiIconText()` using hardcoded enum values

I also found an issue in `GuiTextInputBox()` where the eye toggle was not guarded by `RAYGUI_NO_ICONS`. I added "O" and "*" as placeholders.

The second commit added the macros and replaced all the hardcoded enum values with them.

I tested it on all the examples as well as my own projects and tests. Everything compiled and worked as expected. I did find some issues with `RAYGUI_NO_ICONS` though, but they are unrelated to these changes. I may open a separate issue for that

### DIFF
--- a/src/raygui.h
+++ b/src/raygui.h
@@ -422,6 +422,44 @@
     #define GUI_INPUT_KEY           GetCharPressed()
 #endif
 
+// Macros to define internally used icons
+#if !defined(RAYGUI_ICON_WINDOW_CLOSE)
+    #define RAYGUI_ICON_WINDOW_CLOSE        ICON_CROSS_SMALL
+#endif
+#if !defined(RAYGUI_ICON_TABBAR_CLOSE)
+    #define RAYGUI_ICON_TABBAR_CLOSE        ICON_CROSS_SMALL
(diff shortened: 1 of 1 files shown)


raysan5:
No plans for this change

js-lm:
> No plans for this change

Can the first commit be merged so that I can at least try working with the hard coded values? It also fixed the RAYGUI_NO_ICONS bug in GuiTextInputBox()

raysan5:
> Can the first commit be merged so that I can at least try working with the hard coded values? It also fixed the RAYGUI_NO_ICONS bug in GuiTextInputBox()

@js-lm Ok, can you send a new PR just with those changes, thanks!

Note that I always use the `#134#` approach on all my tools, never the `GuiIconText(GUI_ICON_NAME, "text")`... I see  it generates multiple issues... probably all mechanism should be redesigned... I'll take a closer look in the future...

PR #553, opened 2026-05-17
Replace magic strings with the enum values

js-lm:
Replaced the hardcoded icon ids strings with GuiIconText() using hardcoded enum values

I also found an issue in GuiTextInputBox() where the eye toggle was not guarded by RAYGUI_NO_ICONS. I added "O" and "*" as placeholders.


raysan5:
@js-lm thanks for the review

PR #558, opened 2026-05-30
Working MSVC build command.

KRC2000:
Provided msvc build command did not work, so I fixed it up to make it work, and decided to contribute :)

Hints + linking necessary dependencies on windows: msvcrt.lib winmm.lib gdi32.lib user32.lib shell32.lib.


raysan5:
@KRC2000 thanks for the review

PR #559, opened 2026-05-31
Clearer MSVC build instructions.

KRC2000:
More clarification. MSVC binary build of Raylib provided by official github repository (which is what user will mostly want to link raygui against) actually contains both static and dynamic libraries. And `.lib` file in there is not so called "import library", it's actually a full static library, so linking against it in this case is incorrect. `.lib` extension is confusing, because it could be both a static library and an import library. 

So, user actually wants to link with `raylibdll.lib` and not `raylib.lib`.

In my previous PR I incorrectly hint user into linking `.lib` which can be correct if `.lib` is in fact import library, but in most cases that would not be the case. That is why I am adding this clarification.

<RELATIVE_PATH_TO_raylib.lib> --> <RELATIVE_PATH_TO_RAYLIB_IMPORT_LIBRARY>


raysan5:
@KRC2000 thanks for the review

PR #563, opened 2026-06-28
fix compiler warnings

Vortex2Oblivion:
I was getting compiler warnings from these values not being explicitly casted to `float`s.
First time contributing so let me know if I did something wrong.


raysan5:
It was already fixed.

PR #565, opened 2026-07-11
Fix MSVC compiling with CPP

InviseDivine:
Because MSVC with cpp don't recognize (Type) syntax, we should add RAYGUI_CLITERAL to fix it


raysan5:
@InviseDivine thanks for the review

PR #566, opened 2026-07-15
Change icon ID in README.md example

c-styler:
Updated the prefix from the old RICON to ICON


raysan5:
@som3a-dev thanks for the review!

PR #567, opened 2026-07-28
gcc warnings

mjw6i:
Thank you for your great projects!
I was getting warnings on this line, I hope I got the intent right.


raysan5:
@mjw6i Ouch! Good catch! Thanks for the fix!

PR #568, opened 2026-07-29
fixes textSize parameter shadowing

mglgmz:
GuiTextInputBox method is creating a new textSize variable which shadows the one defined on the GuiTextInputBox method parameters, causing compilers with -Wshadow flag set to emit a warning.

This change renames that method internal variable so it doesn't shadow the function parameters one, completely avoiding the shadow.

Screenshot:
image


raysan5:
@mglgmz thanks for the review!

PR #569, opened 2026-07-31
Fix value clamping logic in raygui.h

666hwll:
This makes a circular spinner possible


raysan5:
@666hwll This changees current functionality, despite it can be interesting in some cases, I prefer to keep current behaviour.

666hwll:
Ok understandable.
Wouldn't a flag or optional parameter circumvent that issue?

PR #571, opened 2026-08-21
add scissor_mode for clipping button clicked outside a rectangle.

wchen1994:
Hello,
this purpose is addressing the issue the gui can be interact outside the scissor rectangle. I think it would be easier if the origin raylib provide a way to get the scissor rectange.
Also check the scroll_panel.c in the example I modified.


raysan5:
@wchen1994 it can be managed at user side, that's what I do on my tools.

wchen1994:
I think it will be better if user doesn't required to modify the raygui's codes. Or can it be other way to achieve without modification. Thank you.

raysan5:
@wchen1994 On user side, use Begin/EndScissorMode()

PR #573, opened 2026-09-06
Add build.zig.zon to raygui

nat3Github:
context: 
zig users can already use raygui just by fetching the source then use whatever they like from it

reason for build.zig.zon:
the .paths declaration in .zon gates what gets copied to the zig-pkg folder. if .paths are not properly gated to whats actually needed, all the additional files like .pngs etc. will get duped into the zig-pkg fodler which leads to unnecessary disk usage.

i am aware this is not really a raygui issue, but as a zig user i would be really happy if this was solved in any way.

related https://github.com/david-vanderson/dvui/pull/975

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,0 +1,19 @@
+.{
+    .name = .raygui,
+    .version = "4.0.0",
+    .minimum_zig_version = "0.16.0",
+
+    .fingerprint = 0x6ef335682ef70c01,
+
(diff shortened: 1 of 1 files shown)


raysan5:
@nat3Github thanks but, diferently than raylib, I prefer to keep build systems as minimal as possible for raygui, no CMake or  Zig (at least  for now).

